PR #108, opened 2026-05-18
[security] supply-chain: Semgrep container pinned by tag rather than digest (Issue #102)

### LINKED ISSUE #102: [security] supply-chain: Semgrep container pinned by tag rather than digest
## Summary

The Semgrep SAST workflow runs in a container pinned by mutable tag (`image: semgrep/semgrep:1.86.0`) rather than by immutable digest. Docker tags are mutable: an attacker who compromises the `semgrep/semgrep` registry namespace could re-push that tag pointing at a malicious image, and the next workflow run would silently execute it.

The Gitleaks workflow demonstrates the correct pattern in the same repository: it downloads the binary and verifies a recorded `sha256` (`.github/workflows/gitleaks.yml:32, 47`). Apply the equivalent here.

## Class

`supply-chain:unpinned-container`

## Severity

**Low** — `AV:N/AC:H/PR:N/UI:N/S:U/C:L/I:L/A:N`. The Semgrep job runs with `permissions: contents: read` and is not granted write tokens; impact is mostly observability (a malicious scanner could fail to report real findings, mask vulnerabilities, or exfiltrate `SEMGREP_APP_TOKEN`). It cannot directly push code.

## Confidence

**high** — the unpinned tag is the only `image:` declaration in the workflow.

Confidence lift: an upstream registry incident showing a `semgrep/semgrep` tag re-push.

## Ease of exploit

**hard** — requires compromise of the upstream registry namespace.

## Attacker model

Attacker who can re-push to the `semgrep/semgrep` Docker Hub namespace.

## Trigger

Any PR triggers the Semgrep workflow, which pulls `semgrep/semgrep:1.86.0` via the runner's container resolution path. If the tag has been re-pushed, the attacker's image is used.

## Why it is a bug

```yaml
# .github/workflows/semgrep.yml:33–35
container:
  # Pinned Semgrep release. Equivalent to `uses: semgrep/semgrep-action@v1`.
  image: semgrep/semgrep:1.86.0
```

The comment claims the image is "pinned" but the tag is mutable. Pinning by tag is informational only; pinning by `@sha256:...` digest is immutable.

## Exploit sketch

1. Attacker compromises the `semgrep/semgrep` Docker Hub namespace.
2. Attacker pushes a malicious image to the `1.86.0` tag.
3. Next PR run pulls the malicious image; the runner executes attacker code with access to `SEMGREP_APP_TOKEN` and the PR's read-only checkout.

## Fix suggestion

Replace the tag pin with a digest pin and record the digest in a comment so bumps are explicit:

```yaml
container:
  # Semgrep v1.86.0, frozen YYYY-MM-DD.
  image: semgrep/semgrep@sha256:<64-hex-digest>
```

Document the digest-bump protocol in the workflow header alongside the existing tag-bump note.

## Files / lines

- `.github/workflows/semgrep.yml` — 30–35

Stable id: `726f6ede`.

Reported by automated security audit per issue #99.

stservice:
# PR Summary — Issue #102

## Summary

Pinned the Semgrep SAST scanner container by immutable `sha256` digest
instead of the mutable `1.86.0` Docker tag, eliminating the silent
tag-repush attack vector flagged in the issue. The validator
(`scripts/check-semgrep-workflow.sh`) is updated to enforce the new
digest-only rule so any future regression to a tag pin fails CI, and
the workflow header now documents the digest-bump protocol alongside
the existing version-label conventions. Closes #102.

## Evidence

This is a CLI/workflow change with no UI surface, so the evidence is
the test results and the resulting workflow YAML.

* `bats tests/scripts/semgrep_workflow.bats` — 16/16 tests pass,
  including two new cases (`fails when the container image is pinned
  to a mutable tag (issue #102)` and `fails when the container image
  digest is malformed`) plus the existing `real repository semgrep
  workflow satisfies every rule` end-to-end gate.
* `./quality.sh` — full local gate passes (shellcheck, all workflow
  validators, every bats suite, cargo-deny, fmt, clippy, check, build,
  test, doc, release).

The new policy enforced by the validator:

```mermaid
flowchart LR
    A["image: semgrep/semgrep<br/>(bare)"] --> X[FAIL: not pinned by digest]
    B["image: semgrep/semgrep:latest"] --> X
    C["image: semgrep/semgrep:1.86.0<br/>(mutable tag)"] --> X
    D["image: semgrep/semgrep@sha256:&lt;64-hex&gt;"] --> Y[PASS]
    E["image: semgrep/semgrep@sha256:short"] --> X
```

Pinned digest (multi-arch manifest, frozen 2026-05-18, looked up via
Docker Hub for the v1.86.0 tag):

```yaml
image: semgrep/semgrep@sha256:a9ea2d5621c29d815d90c2a3b2f9571da8972ef4ff855c9e4902681730240e35
```

## Test Plan

* Updated `tests/scripts/semgrep_workflow.bats`:
  * `write_container_workflow` fixture now emits a digest-pinned image
    so positive cases reflect the new policy.
  * `passes on the container fixture` now asserts on the
    "pinned by digest" OK message.
  * `fails when the container image is unpinned (no tag)` →
    renamed to "no digest" with the same intent.
  * Added `fails when the container image is pinned to a mutable tag
    (issue #102)` — drives the rule introduced by this PR.
  * Added `fails when the container image digest is malformed` —
    guards against an invalid hex digest sneaking past the regex.
  * `:latest` and "no semgrep entry point" cases continue to fail
    as before (untouched intent).
* `real repository semgrep workflow satisfies every rule` exercises
  the actual shipped `.github/workflows/semgrep.yml` against the
  updated validator.

---

🤖 Processed by: stservice

### DIFF
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -10,14 +10,21 @@ name: Semgrep
 #   container path simply runs the CLI directly inside the pinned image
 #   instead of through the action wrapper.
 #
-# Why pin the container tag?
-#   * Reproducibility — a given PR always runs the same scanner version,
+# Why pin the container by digest?
+#   * Reproducibility — a given PR always runs the same scanner build,
 #     so rule updates never silently change PR-gating behaviour.
+#   * Supply-chain hygiene — Docker tags are mutable (an attacker who
+#     compromises the `semgrep/semgrep` namespace could re-push the tag
+#     to point at a malicious image). The `sha256:` digest is immutable
+#     and prevents that silent substitution (Issue #102).
 #   * Transparency — bumping Semgrep is an explicit, reviewable change to
-#     the `image:` tag below.
+#     both the digest and the trailing version-label comment below.
 #
-# Bump protocol: lift the tag and record the upstream changelog highlights
-# in the PR that does the bump.
+# Bump protocol: look up the upstream digest for the new release
+# (`docker buildx imagetools inspect semgrep/semgrep:<version>` or the
+# Docker Hub tag page), update the digest AND the version-label comment
+# in lockstep, and record the upstream changelog highlights in the PR
+# that does the bump.
 
 on:
   pull_request:
@@ -32,7 +39,10 @@ jobs:
     runs-on: ubuntu-latest
     container:
       # Pinned Semgrep release. Equivalent to `uses: semgrep/semgrep-action@v1`.
-      image: semgrep/semgrep:1.86.0
+      # Digest pin: semgrep/semgrep v1.86.0 multi-arch manifest, frozen
+      # 2026-05-18. Docker tags are mutable; the sha256 digest is not
+      # (Issue #102).
+      image: semgrep/semgrep@sha256:a9ea2d5621c29d815d90c2a3b2f9571da8972ef4ff855c9e4902681730240e35
     steps:
       - name: Checkout code
         uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd  # v5
--- a/docs/pr-summary-102.md
+++ b/docs/pr-summary-102.md
@@ -1,0 +1,62 @@
+# PR Summary — Issue #102
+
+## Summary
+
+Pinned the Semgrep SAST scanner container by immutable `sha256` digest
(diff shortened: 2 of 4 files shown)


stservice:

```
external: no updates

```

</details>

Issue #83, opened 2026-05-08
End-to-end GPU benchmark and ship/skip decision for default behaviour


stservice:
## Summary
Run the full benchmark suite (CPU baseline, CPU+PGO, GPU single-creature, GPU multi-creature with pipelining) at the issue-target corpus sizes, update `docs/performance-baseline.md`, and decide whether `--gpu auto` should default to using a detected GPU.

## What Needs to Be Done
1. Run `./scripts/run-benches.sh` with `BENCH_SCORING_BYTES=200000000` against:
   - Current CPU release build.
   - PGO-optimised CPU build (`./scripts/build-pgo.sh`, see Issue #43).
   - GPU build with `--gpu on` (single-creature kernel from #81).
   - GPU build with `--gpu on` (multi-creature batched + pipelined from #82) at N=10 and N=50.
2. Capture median + 95% CI half-width per benchmark, on at least one Apple silicon (Metal) and one Linux+NVIDIA (Vulkan) host. Note the host config in each row.
3. Update `docs/performance-baseline.md` with a new "GPU baseline" section dated and per-host. Append; do not overwrite older sections.
4. Decision logic (codify in `gpu` module):
   - If GPU median is ≥ 3 % faster than CPU+PGO at `BENCH_SCORING_BYTES=200000000` for the given path and the parity tolerance from #81 holds, mark that path as "GPU-default-on" via `GpuMode::Auto`.
   - Otherwise leave that path's `Auto` mode as CPU and document the negative result.
5. If GPU wins on directory mode but loses on single-creature, default `Auto` to GPU only when directory mode is in use; this can be a runtime decision based on the input being a creature dir.
6. Update `README.md`'s "How to bench" and add a new "GPU acceleration" section summarising the supported backends, the `--gpu` flag, the JSON shape change, and the headline numbers.
7. If **no** path shows a meaningful gain, follow the **Performance Task Workflow** negative-result path: post the bench evidence to the parent issue (#78), apply the `negative-result` label, and recommend closing.

## Acceptance Criteria
- [ ] `docs/performance-baseline.md` has a new dated "GPU baseline" section with CPU / CPU+PGO / GPU rows for both Metal and Vulkan hosts.
- [ ] `README.md` has a "GPU acceleration" section.
- [ ] `--gpu auto` selects GPU only on paths where the bench evidence supports it.
- [ ] If applicable, the negative-result path is followed and #78 is updated with the evidence (no PR raised in that branch).
[...]

Issue #81, opened 2026-05-08
GPU-backed fused activation + MSE kernel for forward-only single-creature scoring


stservice:
## Summary
Implement a wgpu compute shader that performs per-record forward-only MLP activation plus fused MSE accumulation, dispatched from `stream_score::accumulate_mse_sum_forward_only_fused` when GPU mode is enabled.

## What Needs to Be Done
1. Add a `shaders/` directory under `rust_scorer/src/` containing:
   - `forward_mse.wgsl` — one workgroup-per-record kernel that:
     - Reads packed `[inputs..., targets...]` records (matching today's CPU memory layout).
     - Runs forward-only activation through a uniform buffer of weights/biases (TANH first; expandable to other squash types in later issues).
     - Computes per-record squared error, reduces into a single `f32` MSE sum via group-shared atomics or a two-pass reduction.
2. Host-side glue in `rust_scorer/src/gpu/forward_mse.rs`:
   - Build the bind-group layout, pipeline, and reusable staging buffers once per scorer run (allocate at startup, reuse per chunk).
   - Per chunk: copy the unpacked `[f32]` records into the GPU input buffer, dispatch, and read back a single `f32` (mse_sum).
   - Avoid round-tripping through CPU floats more than once per chunk — accept the LE-unpacked `Vec<f32>` directly.
3. Wire into `stream_score.rs`:
   - When `GpuMode::On|Auto` and adapter present, use the GPU path; otherwise current CPU `mse_sum_batch_packed` path.
   - JSON output includes `gpuKernel: "forward_mse"` when GPU was used.
4. Parity test: a property-style test runs the same creature + corpus through CPU and GPU and asserts the two MSE sums agree within `1e-4` relative tolerance.
5. Criterion bench: add `gpu_score_from_json_fused/forward_only` to `benches/scoring.rs`. Skip the bench when no GPU adapter is detected so the bench still runs on CI.
6. Document the shader, the dispatch grid sizing, and the parity tolerance in inline doc comments.

## Acceptance Criteria
- [ ] CPU vs GPU MSE sums agree within `1e-4` relative tolerance on the synthetic 8→8→2 fixture used by existing benches.
- [ ] `gpu_score_from_json_fused/forward_only` Criterion bench exists and runs cleanly when a GPU is present.
[...]

PR #183, opened 2026-06-08
Make --gpu auto fall back to CPU cleanly above the 256-neuron shader cap (#180)

### LINKED ISSUE #180: --gpu auto (default) fails batch scoring for creatures above the 256-neuron shader cap; forces slow per-creature CPU scoring
## Summary

With the default `--gpu auto` (default flipped to `auto` in #83), a **batch** scoring call for a creature that exceeds the GPU shader cap (**256 neurons**) **fails** instead of cleanly scoring on CPU. The caller (`@stsoftware/neat-ai`) then abandons batch scoring and drops to **per-creature CPU scoring**, which is dramatically slower. Real evolved creatures are well past this cap (observed: **4139 neurons**), so this happens on essentially every batch call.

The scorer's own error text recommends `rerun with --gpu off`.

## Evidence

From a production learn run (Apple M4 Pro, Metal 4, `--cost MSE`, `--gpu` not passed → default `auto`):

```
[NEAT-AI] Batch rust scorer reconciliation failed (INVALID_JSON):
Rust scorer batch call failed (exit 158): [gpu] auto fallback to CPU directory mode:
GPU runner cannot host this creature set (creature 0 has 4139 neurons;
GPU shader caps at 256); rerun with --gpu off;
Batch scorer rejected 0 creature(s) (forwardOnly=true=19, forwardOnly=false=0): [];
falling back to per-creature scoring.
```

(Several occurrences with 4135–4139 neurons.) Note `exit 158` + invalid JSON on stdout — the batch process exits abnormally rather than returning a valid CPU-scored result.

## Two distinct problems

1. **Broken auto→CPU fallback (bug/regression).** Under `--gpu auto`, when the creature set can't fit the GPU shader, the run should *silently and correctly* fall back to the CPU pipeline and return valid JSON. Instead the batch call fails (exit 158, `INVALID_JSON`), forcing the caller onto a much slower per-creature path. This is the practical regression behind "everything got slow / the learn loop stopped completing" — the message literally says to `rerun with --gpu off`, i.e. `auto` is worse than `off` here.

2. **GPU is effectively unusable for real creatures (shader cap = 256 neurons).** Production creatures are 4000+ neurons, so the GPU path can *never* host them. We have capable GPUs (Apple Silicon / Metal) and want them used — the 256-neuron shader cap means they sit idle while the CPU thrashes. Request: support scoring creatures larger than 256 neurons on the GPU (or document the cap and make the CPU fallback first-class and fast).

## Impact

- Forces slow per-creature CPU scoring on every batch call for large creatures.
- On shared hosts this all-CPU load is severe: multiple concurrent scorer processes (each rayon-parallel across all cores) drove a 12-core box to load ~302.
- It is a contributing trigger to the upstream "evolution never completes / never checks in" hang (filed against `stSoftwareAU/NEAT-AI`): the slow scoring makes training tasks blow their timeout.

## Notes

- `AGENTS.md` still says `off` is the default ("default until #81 lands") — this is **stale**; `rust_scorer/src/main.rs` documents the default as `auto` (since #83), and `README.md` agrees. Worth reconciling the docs.
- Immediate workaround on the caller side: pass `--gpu off` for large-creature workloads.

## Suggested fix

1. Make `--gpu auto` fall back to the CPU pipeline **cleanly** (valid JSON, normal exit) when a creature exceeds the GPU shader cap — never return `exit 158 / INVALID_JSON`.
2. Raise/remove the 256-neuron GPU shader cap (or chunk large creatures) so real creatures can actually use the GPU.

## Environment

- `NEAT-AI-scorer` rust_scorer, Develop (binary built 2026-06-08).
- Apple M4 Pro, Metal 4; `--cost MSE`; default `--gpu auto`.
- Caller: `@stsoftware/neat-ai@5.3.16`.

stservice:
# Make `--gpu auto` fall back to CPU cleanly above the 256-neuron shader cap

## Summary

Under the default `--gpu auto`, **batch** (directory-mode) scoring of a
creature set that exceeds the GPU shader cap (**256 neurons**) created a
`wgpu`/Metal device up-front, then abandoned it when the
`forward_mse_batched` kernel rejected the set. On some drivers (observed:
Apple M4 Pro / Metal 4) that abandoned context aborted during process
teardown, truncating block-buffered stdout — batch callers saw
`exit 158` / `INVALID_JSON` and dropped to dramatically slower per-creature
CPU scoring. Real evolved creatures (4000+ neurons, observed 4139) hit this
on essentially every batch call.

**Fix:** under `--gpu auto` the directory path now runs a **CPU-only
pre-flight** (`multi_score::gpu_directory_compatible`) that checks the
creature set against the shader cap (and the MSE-only / shared-shape
constraints) **before** any GPU adapter is selected. An unhostable set
routes straight to the CPU pipeline without ever creating — or tearing down
— a `wgpu` context, so the fallback returns valid JSON with
`gpuBackend: "cpu-fallback"` and exits 0. `--gpu on` still hard-errors on an
unhostable set (the user explicitly demanded a GPU); small creature sets
still run on the GPU.

This addresses **problem #1** from the issue (broken auto→CPU fallback).
**Problem #2** (raise/remove the 256-neuron cap so large creatures run on
the GPU) is a separate GPU-kernel redesign — the WGSL kernel holds each
creature's activations in a fixed-size `private` array that cannot simply be
enlarged — and is tracked as follow-up issue
[stSoftwareAU/NEAT-AI-scorer#182](https://github.com/stSoftwareAU/NEAT-AI-scorer/issues/182).

Also reconciles the **stale docs** the issue flagged: `AGENTS.md` said `off`
was the default ("until #81 lands") — it is `auto` since #83. The cap and
first-class CPU fallback are now documented in `README.md`.

Closes #180.

## Evidence

Backend/CLI change — no web UI. Verified by building the binary and running
it against a generated creature directory.

**Oversized creature (302 neurons) under `--gpu auto` — clean CPU fallback:**

```text
$ rust_scorer --gpu auto <oversized_creatures_dir> <data_dir> | jq .c0.gpuBackend
exit=0
"cpu-fallback"
# stderr:
[gpu] auto fallback to CPU directory mode: GPU runner cannot host this
creature set (creature 0 has 305 neurons; GPU shader caps at 256); rerun
with --gpu off
```

stdout is valid JSON even when piped (the caller's exact scenario); exit 0.

**No regression — small sets still use the GPU; `--gpu on` still hard-errors:**

```text
$ rust_scorer --gpu auto  <small_creatures_dir> <data_dir>  → gpuBackend "metal"
$ rust_scorer --gpu on    <small_creatures_dir> <data_dir>  → gpuBackend "metal"
$ rust_scorer --gpu on    <oversized_dir>       <data_dir>  → exit 1,
    "Error: GPU runner cannot host this creature set (... 305 neurons ...)"
```

### Control flow (after fix)

```mermaid
flowchart TD
    Mode{GpuMode}
    Mode -->|Off| CPU[CPU pipeline]
    Mode -->|Auto + single creature| CPU
    Mode -->|Auto + directory| Pre{CPU pre-flight #180<br/>set hostable?<br/>≤256 neurons · MSE · shape}
    Mode -->|On| Adapter[wgpu adapter selection]
    Pre -->|no — never makes a GPU device| CPU
    Pre -->|yes| Adapter
    Adapter -->|found| Kernel[forward_mse_batched]
    Adapter -->|none + Auto| CPU
    Adapter -->|none + On| Err[exit non-zero]
    Kernel -->|kernel rejects creature| CPU
```

## Test Plan

New tests (all passing; full workspace suite green):

- `rust_scorer/tests/gpu_preflight_tdd.rs` — drives the new pre-flight
  directly (TDD driver; fails to compile against the unfixed tree because
  the function did not exist):
  - `preflight_rejects_creature_above_shader_cap` — a 302-neuron set returns
    `Err` mentioning the 256 cap.
  - `preflight_accepts_small_creature_set` — a tiny set returns `Ok`.
  - `preflight_defers_empty_directory_to_scoring_path` — load failures are
    *not* reported as GPU-incompatibility (returns `Ok` so the scoring path
    surfaces the real error).
- `rust_scorer/tests/directory_mode_tdd.rs::gpu_auto_directory_above_shader_cap_falls_back_to_cpu_cleanly`
  — binary-level regression guard: `--gpu auto` over an oversized creature
  emits valid JSON with `gpuBackend: "cpu-fallback"` and exit 0.

Existing GPU/auto/cost tests (`scorer_smoke.rs`, `directory_mode_tdd.rs`,
`gpu_multi_score_parity.rs`) continue to pass — no existing tests modified.
`./quality.sh` passes cleanly (fmt, clippy `-D warnings`, cargo-deny,
rustdoc, release build, full test suite).

---

🤖 Processed by: stservice
🏷️ Run id: `vibe-mq4tbwl4-41ae63`


nleck (approved):

nleck:
Why does this consistently fail ? doesn't on thoses like NEAT-AI 

Run set -euo pipefail
Downloading pinned Gitleaks v8.24.2 from https://github.com/gitleaks/gitleaks/releases/download/v8.24.2/gitleaks_8.24.2_linux_x64.tar.gz
curl: (22) The requested URL returned error: 504
Error: Process completed with exit code 22.

stservice:
Claiming PR feedback comment 4646467377 for worker `GRQ-25-70201`.

stservice:
Good question — the failure isn't specific to this PR's code; it's a transient infrastructure error. The `504` comes from GitHub's release-asset CDN while downloading the pinned Gitleaks binary, and the workflow's `curl` had **no retry**, so any one-off gateway blip failed the whole job. Other repos (like NEAT-AI) just happened not to hit the blip on their runs — it's intermittent, not deterministic.

Fix: added retry with backoff to the download step in `.github/workflows/gitleaks.yml`:

- `--retry 5 --retry-delay 5 --retry-all-errors --connect-timeout 30` on the `curl` call.

A transient `504`/timeout is now retried up to 5 times instead of failing immediately. The pinned `SHA256` check is unchanged, so we still only ever accept the expected binary — the retry only affects resilience, not supply-chain integrity.

This is a CI-workflow-only change (no Rust code touched), so the `quality.sh` Rust gates are unaffected.

Issue #199, opened 2026-06-12
Refactor: factor the duplicated `complexity_penalty` formula into one shared scoring function


VibeCoderST:
Part of #198.

## Problem
The `complexity_penalty` arithmetic is hand-copied across multiple call sites purely to populate the `complexityPenalty` JSON field, while `scoring::calculate_score` already computes the identical formula internally:

- `rust_scorer/src/main.rs:451-453`
- `rust_scorer/src/multi_score.rs` (the two result-assembly sites)
- canonical copy in `rust_scorer/src/scoring.rs:184-186`

The blocks are numerically identical today, but any change to the growth-cost weighting (`hidden*GROWTH_COST + synapses*GROWTH_COST/10.0 + total_penalty*GROWTH_COST/100.0`) must be made in every copy or the reported `complexityPenalty` silently disagrees with `score`. The call sites also average raw `value_penalty(...)` while `calculate_score` routes through `calculate_penalty` (which adds finiteness asserts), a subtle inconsistency.

## Suggested fix
Expose a single `pub fn complexity_penalty(components: &ScoreComponents, growth_cost: f64) -> f64` in `scoring.rs`, call it from `calculate_score`, and replace every inline block with that call.

## Acceptance criteria
- One canonical implementation of the complexity-penalty formula.
- All result-assembly sites call it; no duplicated arithmetic remains.
- Existing tests still pass; add a test asserting the JSON `complexityPenalty` equals the value used inside `calculate_score`.

Severity: medium · Effort: low

stservice:
Pull request https://github.com/stSoftwareAU/NEAT-AI-scorer/pull/215 has been created to address this issue.

stservice:
Closed automatically — PR has been merged.

PR #214, opened 2026-06-13
Propagate per-chunk cost errors instead of panicking (Issue #200)

### LINKED ISSUE #200: Robustness: replace `.expect()` in the parallel per-chunk dispatch with error propagation
Part of #198.

## Problem
Inside the Rayon per-chunk dispatch, the cost accumulation result is unwrapped with `.expect("cost validated before per-chunk dispatch")`:

- `rust_scorer/src/multi_score.rs:417`
- mirrored in `rust_scorer/src/stream_score.rs` (per-chunk `expect` sites)

The justification is a one-time empty-chunk probe run before the I/O loop, but that probe passes a zero-length slice, so any per-record failure mode that depends on actual record contents would surface as a panic across a Rayon worker and abort the whole binary — instead of the clean `Err(String)` the rest of the pipeline returns.

## Suggested fix
Propagate the error out of the parallel closure instead of `expect`: capture the first error (e.g. `Mutex<Option<String>>` / `OnceLock`, or collect `Result`s and short-circuit) and return it as `Err` from the scoring function.

## Acceptance criteria
- No `.expect()` on `accumulate_cost_sum(...)` inside parallel loops.
- A malformed-record / failing-cost case returns a structured error rather than panicking.
- Add a regression test that triggers the error path and asserts a clean `Err`.

Severity: medium · Effort: low

VibeCoderST:
## Summary

Replaced the panic-on-error `.expect()` calls on `accumulate_cost_sum(...)` inside the
Rayon per-chunk dispatch with structured error propagation. Previously a per-record
failure mode that depended on actual record contents would surface as a panic across a
Rayon worker and abort the whole binary, instead of the clean `Err(String)` the rest of
the pipeline returns. The one-time empty-chunk probe before the I/O loop only validates a
zero-length slice, so it cannot catch content-dependent failures.

Changes:

- **`rust_scorer/src/cost.rs`** — `accumulate_cost_sum` now rejects a malformed chunk
  whose float length is not a whole multiple of the record stride with a structured
  `Err("Malformed record chunk: ...")`, rather than letting the upstream
  `*_sum_batch_packed` helpers silently truncate a trailing partial record. This makes the
  per-chunk error path reachable and testable.
- **`rust_scorer/src/multi_score.rs`** — the flat `par_iter_mut().for_each(... .expect(...))`
  becomes `try_for_each(...)?`, so any per-worker cost error short-circuits and propagates
  out of the scoring function as `Err`.
- **`rust_scorer/src/stream_score.rs`** — both the fast-path and slow-path parallel
  reductions now `collect::<Result<Vec<f64>, String>>()` and `?`, and the single-network
  arms use `?` instead of `.expect`. The up-front probe comment was updated accordingly.

Closes #200.

## Evidence

This is a backend/CLI change with no web interface to screenshot. Verified via the Rust
test suite (see Test Plan). The control-flow change for the parallel dispatch:

```mermaid
flowchart TD
    A[per-chunk dispatch] --> B{accumulate_cost_sum per worker}
    B -->|Ok sum| C[reduce into per-creature totals]
    B -->|Err| D["before: .expect → panic, aborts binary"]
    B -->|Err| E["after: ? / collect Result → clean Err(String) returned"]
    style D stroke:#c0392b
    style E stroke:#27ae60
```

Test output (lib tests):

```text
test cost::tests::accumulate_cost_sum_rejects_malformed_chunk ... ok
test cost::tests::accumulate_cost_sum_error_propagates_through_rayon ... ok
test result: ok. 63 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

Note: the integration test `gpu_auto_directory_above_shader_cap_falls_back_to_cpu_cleanly`
fails on the local Apple-Metal dev machine (it observes `metal` rather than `cpu-fallback`).
This failure is pre-existing and environment-specific — it reproduces identically on a
clean checkout without this change — and is unrelated to this issue. CI runs on Linux
without a Metal adapter and uses the CPU fallback path.

## Test Plan

Added two regression tests in `rust_scorer/src/cost.rs`:

- `accumulate_cost_sum_rejects_malformed_chunk` — a chunk whose float length is not a whole
  multiple of the record stride returns a structured `Err` (no panic, no silent truncation).
- `accumulate_cost_sum_error_propagates_through_rayon` — replicates the production
  `par_iter_mut().map(...).collect::<Result<_, _>>()` pattern with one malformed slice and
  asserts the error surfaces as a clean `Err` rather than panicking across a Rayon worker.

All `rust_scorer` library tests pass (65 lib + 86 integration). `cargo fmt --all --check`,
`cargo clippy --all-targets` and `RUSTDOCFLAGS="-D warnings" cargo doc` are clean.

## Merge / base reconciliation

This branch was rebased onto the latest `milestone/improvements` (which now includes #199 and
the auto version bump). Conflict resolution:

- `rust_scorer/Cargo.toml` — kept the auto-bumped `0.5.54`.
- `rust_scorer/src/scoring.rs` — the merged-in #199 doc comment linked the public
  `complexity_penalty` to the **private** `calculate_penalty`, which fails
  `RUSTDOCFLAGS="-D warnings" cargo doc` (`rustdoc::private_intra_doc_links`). Converted the
  intra-doc link to a plain code span so the rustdoc gate passes. No behaviour change.

## Milestone
This PR is part of the **Improvements** milestone and targets the `milestone/improvements` feature branch.

---

🤖 Processed by: stservice
🏷️ Run id: `vibe-mqbqx3gz-f3a387`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -648,13 +648,12 @@ dependencies = [
 
 [[package]]
 name = "js-sys"
-version = "0.3.99"
+version = "0.3.102"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "142bc4740e452c1e57ade0cbc129f139c9093e354346f0872ef985f4f5cf5f11"
+checksum = "03d04c30968dffe80775bd4d7fb676131cd04a1fb46d2686dbffbaec2d9dfd31"
 dependencies = [
  "cfg-if",
  "futures-util",
- "once_cell",
  "wasm-bindgen",
 ]
 
@@ -773,7 +772,7 @@ dependencies = [
 
 [[package]]
 name = "neat-core"
-version = "0.1.38"
+version = "0.1.39"
 dependencies = [
  "serde",
  "serde_json",
@@ -1109,7 +1108,7 @@ checksum = "19b30a45b0cd0bcca8037f3d0dc3421eaf95327a17cad11964fb8179b4fc4832"
 
 [[package]]
 name = "rust_scorer"
-version = "0.5.51"
+version = "0.5.54"
 dependencies = [
  "bytemuck",
  "clap",
@@ -1383,9 +1382,9 @@ dependencies = [
 
 [[package]]
 name = "wasm-bindgen"
-version = "0.2.122"
+version = "0.2.125"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3ed04576f974d2b2fba0f38c51dbc5518011e38c36bf1143164be765528fd409"
+checksum = "8ddb3f79143bced6de84270411622a2699cee572fc0875aeaf1e7867cf9fca1a"
 dependencies = [
  "cfg-if",
  "once_cell",
@@ -1396,9 +1395,9 @@ dependencies = [
 
 [[package]]
 name = "wasm-bindgen-futures"
-version = "0.4.72"
+version = "0.4.75"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9473dbd2991ae90b6291c3c32c30c6187ac49aa32f9905d1cce280ec1e110b0f"
+checksum = "503b14d284f2c8dac03b819967e155ea753f573586193b2b2c95990cb5d69280"
 dependencies = [
  "js-sys",
  "wasm-bindgen",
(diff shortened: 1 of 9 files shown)


nleck (approved):

stservice:
Rebased/merged the latest `milestone/improvements` into this branch to clear the merge conflict (it predated #199 and the auto version bump). Resolution: kept the auto-bumped `0.5.54` in `Cargo.toml`, and converted the #199 `scoring.rs` doc comment's private intra-doc link (`complexity_penalty` → private `calculate_penalty`) to a plain code span so the `RUSTDOCFLAGS="-D warnings" cargo doc` gate passes. No behaviour change. The per-chunk `.expect()` → error-propagation work and its regression tests are unchanged; `fmt`/`clippy`/`doc`/tests are green locally (the only local test failure is the pre-existing Apple-Metal `gpu_auto_directory_above_shader_cap_falls_back_to_cpu_cleanly`, which passes on GPU-less CI).

Issue #177, opened 2026-06-07
🟢 Add `SECURITY.md` with a vulnerability disclosure policy


VibeCoderST:
None of the three locations GitHub recognises for a security policy — `SECURITY.md` (repo root), `.github/SECURITY.md`, or `docs/SECURITY.md` — exists in this repository. This is a `severity:low` finding in the `general` (repo-level hygiene) bucket. It is `low` rather than `medium` because the repo has no external-publish CI workflow (a `cargo publish` is infeasible while `neat-core` is a `path` dependency), so it reads as an internal / single-consumer tool.

## Why this matters
`SECURITY.md` is the canonical, GitHub-recognised place to declare how to report a vulnerability **privately**, the expected response time, and which versions are supported. Without it, a reporter has no private channel and defaults to opening a public issue — which discloses the vulnerability before a fix exists. GitHub surfaces the file in the repository's Security tab as a community-health file.

## Suggested fix
Add `SECURITY.md` (root or `.github/`) declaring:

- a private reporting route — GitHub private vulnerability reporting (Settings → Security → enable) or a dedicated security contact email;
- the expected acknowledgement / response time;
- a supported-versions table.

GitHub's guide is the canonical reference: <https://docs.github.com/en/code-security/getting-started/adding-a-security-policy-to-your-repository>.

VibeCoderST:
Pull request https://github.com/stSoftwareAU/NEAT-AI-scorer/pull/186 has been created to address this issue.

PR #276, opened 2026-07-02
Return Result from GPU score_chunk so readback failures fall back to CPU (Issue #273)

### LINKED ISSUE #273: 🟢 GPU readback map_async failure panics instead of falling back to CPU in rust_scorer/src/gpu/forward_mse_batched.rs
A `severity:low` `lang:rust` best-practices finding in
`rust_scorer/src/gpu/forward_mse_batched.rs:681-684`.

`BatchedRunner::score_chunk` maps the GPU readback buffer and unwraps the
`map_async` result with two `expect()` calls:

```rust
receiver
    .recv()
    .expect("partials map_async sender dropped")
    .expect("partials map_async failed");
```

`score_chunk` returns `Vec<f64>` (not `Result`), so a `map_async` failure at
readback time panics the whole process rather than surfacing an error the
caller can recover from. This is a panic on **external I/O** (the GPU device),
which the `rust` bucket's error-handling check (`prompts/best_practices/buckets/rust.md`,
check 1) flags: prefer `?`/`Result` over `expect()` on external I/O.

## Why this matters

The crate is deliberately designed to fall back to the CPU when the GPU path
fails. `score_from_creature_dir_gpu` (`rust_scorer/src/multi_score.rs:514`,
which calls `score_chunk` at `multi_score.rs:612` and `:651`) returns
`Result<_, String>`, and the `--gpu auto` caller in `main.rs:296-312` catches
that `Err` and re-runs on the CPU:

```rust
Err(e) => {
    eprintln!("[gpu] auto fallback to CPU directory mode: {e}");
    // ... CPU fallback ...
}
```

A mid-run `map_async` failure — a recoverable device hiccup such as device
loss, an out-of-memory condition, or a validation error during readback —
bypasses that design: because `score_chunk` panics instead of returning `Err`,
the panic unwinds past the `Result` boundary and aborts the process under the
default `--gpu auto` mode, turning a recoverable GPU error into a crash. The
two adjacent `u32::try_from(...).expect("... exceeds u32::MAX")` calls
(lines 543-544, 556) are **not** in scope — `n_records` is bounded by the
≤64 MiB read buffer, so those branches are genuinely unreachable.

Severity is `low`: the trigger is a device-failure path (not user input), it
affects only the GPU directory path, and it requires an actual GPU fault to
manifest.

## Suggested fix

Thread the readback failure through as an error so the existing `--gpu auto`
CPU fallback can absorb it, rather than panicking:

- Change `BatchedRunner::score_chunk` to return
  `Result<Vec<f64>, String>` and replace the `expect()` pair at
  lines 681-684 with `?`-propagated errors, e.g.
  `receiver.recv().map_err(|_| "partials map_async sender dropped".to_string())?
  .map_err(|e| format!("partials map_async failed: {e:?}"))?;`
- Propagate the `Result` up through the `score_chunk` call sites in
  `multi_score.rs:612` and `:651` (both already inside a function that returns
  `Result<_, String>`), so a readback failure returns `Err` and lets
  `main.rs`'s `--gpu auto` path fall back to the CPU instead of aborting.

This keeps the panic-free, CPU-fallback guarantee consistent for the readback
failure mode, matching the error handling already applied to GPU
initialisation and preparation (`GpuInitError` / `GpuPrepareError`).

🏷️ Filed by idle-task template: `best-practices` · Run id: `vibe-mr32hmko-f60f87`

VibeCoderST:
## Summary

`BatchedRunner::score_chunk` (`rust_scorer/src/gpu/forward_mse_batched.rs`)
mapped the GPU readback buffer and unwrapped the `map_async` result with two
`expect()` calls. Because `score_chunk` returned `Vec<f64>` (not `Result`), a
`map_async` failure at readback time — a recoverable device fault such as
device loss, an out-of-memory condition, or a validation error — panicked and
unwound past the `Result` boundary, aborting the whole process under the
default `--gpu auto` mode instead of falling back to the CPU.

This change threads the readback failure through as an error so the existing
`--gpu auto` CPU fallback in `main.rs` can absorb it:

- `score_chunk` now returns `Result<Vec<f64>, String>`. The `expect()` pair is
  replaced with a small `interpret_map_async` helper that converts the
  two-level `map_async` channel result into a flat `Result<(), String>`,
  `?`-propagated at the call site; the empty-chunk short-circuit and success
  path return `Ok(...)`. The helper is generic over the error type so the
  error-threading logic is unit-testable **without a GPU** (see Test Plan).
- Both call sites in `multi_score.rs` propagate the `Result`:
  - the synchronous path uses `?` (surfacing through `run_io_loop` to the
    `Result<_, String>` returned by `score_from_creature_dir_gpu`);
  - the pipelined worker forwards the `Err` through its `result_tx` channel,
    which the existing `drain_one` loop already maps to `Err`.

The two adjacent `u32::try_from(...).expect(...)` calls were left untouched —
per the issue they are genuinely unreachable (bounded by the ≤64 MiB read
buffer) and out of scope.

Closes #273

## Evidence

Backend/CLI change with no web interface — no screenshot applicable. Verified
via `cargo test -p rust_scorer` (all tests pass) plus `cargo fmt --check` and
`cargo clippy --all-targets -- -D warnings` (both clean).

The new GPU-gated test ran on this machine's Metal device (not skipped) and
confirmed the healthy-device readback returns `Ok`

Error-propagation flow on a readback failure:

```mermaid
flowchart LR
    A[map_async fails] --> B[score_chunk returns Err]
    B --> C[multi_score call site propagates Err]
    C --> D[score_from_creature_dir_gpu returns Err]
    D --> E["main.rs --gpu auto catches Err"]
    E --> F[re-run on CPU]
```

## Test Plan

- Added three CI-runnable unit tests in
  `rust_scorer/src/gpu/forward_mse_batched.rs` for `interpret_map_async` — the
  pure core of the readback error handling — which run **without a GPU** and
  close the coverage gap left by the GPU-gated happy-path test:
  - `interpret_map_async_ok_passes_through` — `Ok(Ok(()))` yields `Ok(())`;
  - `interpret_map_async_dropped_sender_is_error_not_panic` — a dropped sender
    yields `Err` containing "sender dropped" (no panic);
  - `interpret_map_async_failed_readback_is_error_not_panic` — a failed
    `map_async` yields `Err` containing "map_async failed" and preserves the
    underlying cause.
- Added `rust_scorer/tests/gpu_multi_score_parity.rs::score_chunk_returns_ok_on_healthy_device`
  — GPU-gated (skips cleanly with no adapter); asserts the empty-chunk
  short-circuit and a real dispatch both return `Ok` on a healthy device,
  documenting the new `Result` contract.
- Updated the existing parity test to `.expect(...)` the `Result` from
  `score_chunk`.
- Existing `rust_scorer/src/stream_io.rs::run_io_loop_propagates_score_chunk_error`
  continues to cover propagation of a `score_chunk` error up through
  `run_io_loop` (the synchronous path).

## Notes

`quality.sh` currently stops at a **pre-existing, unrelated** gate:
`breaking neat-core bump: 0.2.2 exceeds handled baseline 0.1.46`. That gate
reflects sibling `NEAT-AI-core` drift and is out of scope for this issue
(clearing it is a separate deliberate PR per the README "neat-core
breaking-bump gate"). The Rust checks this change touches — `fmt`, `clippy`,
and `cargo test -p rust_scorer` — all pass.

---

🤖 Processed by: stservice
🏷️ Run id: `vibe-mr42tk4c-88a03a`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -223,6 +223,7 @@ jobs:
             "rust_scorer/src/main.rs"
             ".github/CODEOWNERS"
             "SECURITY.md"
+            "neat-core.expected-version"
           )
           missing_files=()
           for file in "${required_files[@]}"; do
@@ -237,6 +238,16 @@ jobs:
 
       - name: Validate Cargo workspace
         run: cargo metadata --no-deps --format-version 1 > /dev/null
+
+      # Gate against an unhandled breaking neat-core bump (Issue #252). scorer
+      # keeps the unpinned `path` dependency that tracks head, so this step
+      # fails the build when the sibling neat-core's breaking component (major
+      # for >= 1.0, minor for pre-1.0) climbs above the recorded baseline in
+      # neat-core.expected-version — forcing a deliberate upgrade rather than
+      # tracking head blindly. Reuses the neat-core checkout + sibling symlink
+      # above; the default manifest path resolves via that symlink.
+      - name: Gate on unhandled breaking neat-core bump
+        run: ./scripts/check-neat-core-version.sh
 
   shell-checks:
     name: Shell Script Quality
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,18 @@ section to the released version with its date.
 
 ### Added
 
+- **CI gate against an unhandled breaking neat-core bump (Issue #252).** The
+  `neat-core` dependency stays an unpinned `path` dependency that tracks head
+  (kept by design), so a new version-baseline gate now guards against silently
+  consuming a breaking change. Scorer records the last-handled neat-core
+  version in the checked-in `neat-core.expected-version` file;
+  `scripts/check-neat-core-version.sh` (run in the CI `validation` job and
+  locally via `./quality.sh`) reads the sibling
+  `../NEAT-AI-core/Cargo.toml` `[workspace.package] version` and **fails**
+  when neat-core's breaking component (major for `>= 1.0`, minor for pre-1.0)
+  exceeds the baseline — forcing a deliberate upgrade. Documented in the
+  README and CONTRIBUTING; covered by `tests/scripts/neat_core_version_gate.bats`.
+
 - **stderr note when a non-MSE `--cost` forces CPU fallback in directory
   mode (Issue #205).** Under the default `--gpu auto`, selecting a non-MSE
   `--cost` makes `auto_should_use_gpu` return false, so the directory path
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,6 +23,16 @@ parent/
   NEAT-AI-scorer/    # this repository
 ```
 
+The `neat-core` path dependency is **unpinned** and tracks head, so a
+**breaking** neat-core change can reach scorer silently. CI guards against
+this with the **neat-core breaking-bump gate** (`scripts/check-neat-core-version.sh`):
+it fails when neat-core's breaking component (major for `>= 1.0`, minor for
+pre-1.0) climbs above the version recorded in
+[`neat-core.expected-version`](./neat-core.expected-version). When the gate
+fails, update `rust_scorer` for the breaking change and bump that baseline
+file in the same PR. See the README "neat-core breaking-bump gate" section
+for the full rationale.
+
 ## Prerequisites
 
 The local gate and CI expect the following tools on your `PATH`:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -92,16 +92,10 @@ dependencies = [
 ]
 
 [[package]]
-name = "anyhow"
-version = "1.0.102"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
-
-[[package]]
 name = "arrayvec"
-version = "0.7.6"
+version = "0.7.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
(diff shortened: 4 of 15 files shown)


VibeCoderST:
## CI fix — Project Validation (breaking neat-core bump gate)

**Root cause:** The `Project Validation` job failed at the **Gate on unhandled breaking neat-core bump** step (`scripts/check-neat-core-version.sh`), not at the file-existence check. The sibling `neat-core` is now at **0.2.2** while the recorded baseline in `neat-core.expected-version` was **0.1.46**. Pre-1.0, a higher minor reads as breaking, so the gate failed:

```
FAIL: breaking neat-core bump: 0.2.2 exceeds handled baseline 0.1.46 (pre-1.0 minor increased)
```

**Why a baseline bump is the correct fix:** The breaking change behind the 0.1→0.2 bump is neat-core #177 (`SynapseData::from_index` narrowed u32→u16). rust_scorer **already handles** it — `rust_scorer/src/gpu/forward_mse_batched.rs` widens `u16→u32` at the GPU/WGSL boundary, with regression test `build_batched_network_data_widens_from_index_to_u32`. The 0.2.1/0.2.2 releases are patch drift. So the code side was already done; the only outstanding action per the gate's own remediation was recording the handled baseline.

**Change:** `neat-core.expected-version` `0.1.46` → `0.2.2`.

**Verification (local):**
- `cargo check --workspace` — compiles cleanly against neat-core 0.2.2.
- `cargo test -p rust_scorer --lib widens_from_index` — passes.
- `./scripts/check-neat-core-version.sh` — now exits 0: `OK neat-core 0.2.2 matches handled baseline 0.2.2`.

Note: the `Shell Script Quality` and `CI Required Checks` jobs also showed failures on the same run; `CI Required Checks` is the aggregate gate that fails when any gating job fails, so it should clear once the jobs re-run green. If `Shell Script Quality` is still red after re-run it is a separate issue from this validation gate.

nleck (approved):

Issue #42, opened 2026-04-25
Eliminate per-worker creature recompilation in multi_score


stservice:
## Summary
`multi_score.rs` calls `compile_creature` once per (creature × worker) pair (see lines 240–253). For 50 creatures × 16 workers that is 800 compiles before any scoring runs. Investigate avoiding the duplicate compilation.

## What Needs to Be Done
1. **Investigate the upstream blocker.** The code comment in `stream_score.rs` references `NEAT-AI-core#11` (`CompiledNetwork: Clone` not yet landed). Open or check the upstream issue; if landing it is feasible, do that first and bump the `neat-core` path dep.
2. **If upstream is blocked**, implement a local workaround in `rust_scorer`:
   - Compile the creature once.
   - Identify which `CompiledNetwork` fields are read-only (topology, weights, biases) versus per-activation scratch (activation/hint/trace buffers).
   - Manually clone the read-only state and allocate fresh scratch buffers per worker — this is essentially what NEAT-AI-core#11 will do, just kept in `rust_scorer` until upstream lands.
3. Either way, capture the time spent in `compile_creature` separately from scoring time (add a `compileTimeSecs` field to `ScoreResult` if useful for diagnostics).
4. Bench `score_from_creature_dir` for population sizes 10, 50, 200 with `activation_threads = available_parallelism()`. Compare before/after.

## Acceptance Criteria
- [ ] Either NEAT-AI-core#11 has landed and the bump is in `Cargo.lock`, OR a local workaround compiles each creature exactly once regardless of `activation_threads`.
- [ ] Bench shows reduced startup time for population ≥ 50 (or the issue closes as `negative-result`).
- [ ] All existing tests still pass.
- [ ] `quality.sh` passes.

## Dependencies
Depends on #36

## Context
Part of #35.

This is the largest fixed-cost in directory mode for large populations. For a 200-creature population on a 16-CPU host, the current code does 3 200 `compile_creature` calls before scoring even starts. Eliminating the duplicates is straightforward once the upstream `Clone` impl exists; the only risk is correctness of the manual clone, which the smoke tests catch.

stservice:
Automatic recovery: this issue was unassigned from `stservice` because the worker has no active heartbeat for this issue (assigned over 30 minutes ago with no heartbeat recorded). This likely means the worker crashed between claiming the issue and starting work. The issue is now available to be picked up again.

---

Assigned-without-heartbeat recovery (Issue #632)

[...]

PR #361, opened 2026-07-18
Replace flaky compileTimeSecs wall-clock assertion with compile-count probe (Issue #355)

### LINKED ISSUE #355: 🟡 Wall-clock threshold assertion on `compileTimeSecs` in rust_scorer/tests/directory_mode_tdd.rs:320
`rust_scorer/tests/directory_mode_tdd.rs:317-328` (test `directory_mode_emits_compile_time_secs`, Issue #42 regression guard) asserts a wall-clock budget — `assert!(compile_secs < 1.0, …)` — on the `compileTimeSecs` value reported in directory-mode JSON output. Severity: medium. This is the only timing threshold in the suite.

## Why this matters

This is a performance/timing assertion inside a unit/integration test (audit check: timing assertions). Wall-clock thresholds are flaky on a saturated CI runner or under slow instrumented builds, and provide no reliable signal across hosts (ARM vs x86, loaded vs idle). Worse, this particular cap is also toothless against the regression it documents: the comment says the guard exists so a per-worker recompile "will balloon well past the cap", but for the two tiny fixture creatures even 32 sub-millisecond recompiles would still land far under 1.0 s — so the assertion can flake without a regression and pass with one. A behaviour-based (WHAT) test asserts the invariant the spec requires ("each creature is compiled exactly once"), not a machine-dependent duration the current implementation happens to achieve.

## Suggested fix

Two valid resolutions:

- **(a) Rewrite** the assertion to express the real invariant behaviourally — e.g. a compile-count probe analogous to the existing `training_pass_probe` in `rust_scorer/src/multi_score.rs` (reset, score a directory of N creatures, assert the observed compile count equals N, not N × workers). The existing `compile_secs >= 0.0` presence/shape assertions on the JSON contract can stay.
- **(b) Delete** the `compile_secs < 1.0` upper-bound assertion — deleting a counter-productive test (or assertion) is an acceptable PR outcome. Performance budgets belong in a dedicated benchmark (`rust_scorer/benches/`) with reported numbers, not in the test runner.

🏷️ Filed by idle-task template: `test-audit` · Run id: `vibe-mrpgqksa-3fb59d`

stservice:
## Summary

Replaced the flaky wall-clock threshold assertion (`compileTimeSecs < 1.0`) in
`rust_scorer/tests/directory_mode_tdd.rs` with a behavioural (WHAT) guard on the
real invariant it was meant to protect: directory-mode scoring compiles each
creature **exactly once** (Issue #42), not once per (creature × worker).

The old upper bound was both flaky and toothless — it could fail on a saturated
CI runner without any regression, yet 32 sub-millisecond recompiles of the tiny
fixtures would still land far under 1.0 s, so it could also pass *with* the
regression. Wall-clock budgets belong in a benchmark, not the test runner.

This PR takes both remedies the audit suggested: it **deletes** the toothless
upper bound from the subprocess contract test (keeping the `compileTimeSecs >=
0.0` JSON shape assertion) and **adds** a behavioural compile-count probe that
asserts the invariant directly.

Closes #355.

## What changed

- `rust_scorer/src/multi_score.rs`: added a process-global `compile_probe`
  module (`reset` / `count` / `record_compile`), mirroring the existing
  `training_pass_probe`. `record_compile()` is called at each `compile_creature`
  site in the **batch scoring paths** (CPU and GPU) — not the CPU-only
  pre-flight hostability/topology probes, which are not part of the scoring pass.
- `rust_scorer/tests/directory_mode_tdd.rs`: removed the `compile_secs < 1.0`
  wall-clock assertion; kept the non-negative JSON-shape assertion; updated the
  doc comment to point at the new behavioural test.
- `rust_scorer/tests/compile_once_assertion.rs` (new): resets the probe, scores
  a directory of N creatures on the CPU path, and asserts the observed compile
  count equals N. The pre-#42 per-worker recompile would report `N × workers`
  on any multi-core host and fail the `assert_eq!`.

```mermaid
flowchart LR
    A[score N creatures] --> B[compile_probe::reset]
    B --> C[compile each creature once]
    C -->|clone CompiledNetwork\nfor extra workers| D[score batch in one pass]
    C -.record_compile() ×N.-> E[compile_probe::count == N]
```

## Evidence

Backend/CLI change — no web interface to screenshot. Verified via the test
suite.

Affected tests (all green under `cargo test` and `./quality.sh`):

- `compile_once_assertion::multi_creature_batch_compiles_each_creature_once`
  (N ∈ {2, 5, 11}) — new behavioural guard reproducing the Issue #42 invariant.
- `compile_once_assertion::single_creature_compiles_once`.
- `directory_mode_tdd::directory_mode_emits_compile_time_secs` — JSON contract
  still enforced, minus the flaky wall-clock cap.
- `single_pass_assertion::*` — unchanged, still green (the sibling probe pattern
  this change mirrors).

## Test Plan

- `cargo test -p rust_scorer --test compile_once_assertion --test
  directory_mode_tdd --test single_pass_assertion` — all pass.
- `./quality.sh` — full gate passes (fmt, clippy, check, build, test, doc,
  release).

---

🤖 Processed by: stservice
🏷️ Run id: `vibe-mrpmb6v1-5211cf`

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -54,6 +54,17 @@ jobs:
         # node20, and no v2.1 / v3 tag has shipped yet, so we SHA-pin to the
         # master commit. Issue #136 cleared the prior Node 20 exception.
         uses: rustsec/audit-check@858dc40f52ca2b8570b7a997c1c4e35c6fc9a432  # v2 (master HEAD, Node 24, post upstream #48)
+        # This action compiles cargo-audit from source with `cargo install`.
+        # `rust-toolchain.toml` pins the repo to a concrete rustc for
+        # reproducible builds of *our* code, but that pin must not gate the
+        # build of a third-party *tool*: a transitive cargo-audit dep can raise
+        # its MSRV above our pin and break this step with no code change (e.g.
+        # kstring@2.0.4 requires rustc 1.96.0, our pin is older — PR #361).
+        # RUSTUP_TOOLCHAIN overrides rust-toolchain.toml (highest rustup
+        # precedence) so cargo-audit builds under the installed `stable`
+        # toolchain, decoupling the tool's build from our code's MSRV pin.
+        env:
+          RUSTUP_TOOLCHAIN: stable
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
 
--- a/docs/archive/pr-summaries/pr-summary-355.md
+++ b/docs/archive/pr-summaries/pr-summary-355.md
@@ -1,0 +1,63 @@
+## Summary
+
+Replaced the flaky wall-clock threshold assertion (`compileTimeSecs < 1.0`) in
+`rust_scorer/tests/directory_mode_tdd.rs` with a behavioural (WHAT) guard on the
+real invariant it was meant to protect: directory-mode scoring compiles each
+creature **exactly once** (Issue #42), not once per (creature × worker).
+
+The old upper bound was both flaky and toothless — it could fail on a saturated
+CI runner without any regression, yet 32 sub-millisecond recompiles of the tiny
+fixtures would still land far under 1.0 s, so it could also pass *with* the
+regression. Wall-clock budgets belong in a benchmark, not the test runner.
+
+This PR takes both remedies the audit suggested: it **deletes** the toothless
+upper bound from the subprocess contract test (keeping the `compileTimeSecs >=
+0.0` JSON shape assertion) and **adds** a behavioural compile-count probe that
+asserts the invariant directly.
+
+Closes #355.
+
+## What changed
+
+- `rust_scorer/src/multi_score.rs`: added a process-global `compile_probe`
+  module (`reset` / `count` / `record_compile`), mirroring the existing
+  `training_pass_probe`. `record_compile()` is called at each `compile_creature`
+  site in the **batch scoring paths** (CPU and GPU) — not the CPU-only
+  pre-flight hostability/topology probes, which are not part of the scoring pass.
+- `rust_scorer/tests/directory_mode_tdd.rs`: removed the `compile_secs < 1.0`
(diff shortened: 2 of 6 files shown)


nleck (approved):

stservice:
## CI fix — `security / security` (cargo exit 101)

**Root cause.** The `security / security` job failed at step 6, the
`rustsec/audit-check` action — not at the advisory scan itself. That action
compiles `cargo-audit` from source via `cargo install`. The repo's
`rust-toolchain.toml` pins rustc to `1.95.0`, which (per rustup precedence)
overrides the `stable` toolchain the job installs. A transitive `cargo-audit`
dependency, `kstring@2.0.4`, now requires rustc `1.96.0`, so the build aborts:

```
error: failed to compile `cargo-audit v0.22.2`
Caused by:
  rustc 1.95.0 is not supported by the following package:
    kstring@2.0.4 requires rustc 1.96.0
```

The `##[error]... cargo failed with exit code 101` in the annotation is this
compile failure. The belt-and-braces `Run cargo audit` step and the standalone
`audit` job both pass because they fetch a **prebuilt** cargo-audit binary
(`taiki-e/install-action`) and never compile it.

**Fix.** The toolchain pin exists to reproducibly build and lint *our own code*
— it should not gate the build of a third-party *tool*, otherwise any
cargo-audit dependency raising its MSRV breaks CI with no code change. I set
`RUSTUP_TOOLCHAIN: stable` on the `rustsec/audit-check` step so cargo-audit
builds under the installed `stable` toolchain (1.97.1 in CI, which satisfies
kstring's 1.96.0 floor), decoupling the tool build from our MSRV pin. The
advisory scan and the action's PR annotations are otherwise unchanged.

- File: `.github/workflows/security.yml`
- Validated locally with `actionlint` (clean) and a YAML parse check.
- No Rust source or lockfile changed, so the repo quality gate is unaffected by
  this workflow-only change.

PR #413, opened 2026-07-21
Disable checkout credential persistence in security.yml (Issue #388)

### LINKED ISSUE #388: 🟠 Job `security` checkout persists credentials (`.github/workflows/security.yml`)
**File:** `.github/workflows/security.yml`:34
**Severity:** medium

## Why this matters

Job `security` step 1 runs `actions/checkout` without `persist-credentials: false`. By default checkout writes the workflow's `GITHUB_TOKEN` into `.git/config` as an auth header, where any later step in the job — including a compromised dependency or an injected script — can read it and act as the token. This job shows no static sign of pushing back to the repository or fetching private submodules, so it does not need the persisted credential, and keeping it on disk only widens the blast radius of a compromised step.

## Suggested fix

Add `persist-credentials: false` to the checkout step so the token is not written to disk:

```yaml
      - uses: actions/checkout@<sha>
        with:
          persist-credentials: false
```

If a later step in this job genuinely pushes back to the repository (or fetches a private submodule) using the checkout credential, this is a false positive — suppress it with an in-source `# best-practice-ignore: BP-PERSIST-CREDS-security-security-1 — <reason>` comment above the `uses:` line.

## Evidence

`.github/workflows/security.yml`:34 — `uses: actions/checkout` with no `persist-credentials: false`

stservice:
# Disable checkout credential persistence in `security.yml` (Issue #388)

## Summary

The `security` reusable workflow ran both `actions/checkout` steps with the
default `persist-credentials: true`, writing the workflow's `GITHUB_TOKEN` into
`.git/config` as an auth header. Any later step in the job — a compromised
dependency, an injected script — could read that token and act as it. The job
only reads the checked-out code and runs `cargo audit` / dependency-review; it
never pushes back and never fetches a private submodule, so the persisted
credential was pure blast radius.

This PR adds `persist-credentials: false` to both checkout steps, plus a
`quality.sh`-wired guard so the property cannot silently regress. Closes #388.

## Changes

- `.github/workflows/security.yml` — `persist-credentials: false` on both the
  main checkout and the NEAT-AI-core sibling checkout.
- `scripts/check-persist-credentials.sh` — new indentation-aware YAML scanner
  (same house style as `check-ci-permissions.sh`) that fails when any
  `actions/checkout` step in the target workflow omits `persist-credentials:
  false`. A documented `# best-practice-ignore: BP-PERSIST-CREDS — <reason>`
  comment above the `uses:` line is honoured as an explicit exception.
- `quality.sh` — runs the new check alongside the other workflow-hygiene gates.
- `README.md` — new "Checkout credential persistence" subsection documenting the
  rule and its escape hatch.

```mermaid
flowchart LR
    A[checkout default] -->|token written to .git/config| B[later step reads it]
    B --> C[acts as GITHUB_TOKEN]
    D[persist-credentials: false] -->|token off disk| E[later step has nothing to steal]
```

## Evidence

Backend/CI change — no web interface to screenshot. Verification is the new
check script plus its BATS suite.

`scripts/check-persist-credentials.sh` against the fixed workflow:

```
OK   .../security.yml: checkout at line 23 sets persist-credentials: false
OK   .../security.yml: checkout at line 38 sets persist-credentials: false
```

`shellcheck` (clean) and `actionlint .github/workflows/security.yml` (clean)
both pass.

## Test Plan

- Added `tests/scripts/persist_credentials.bats` (8 tests):
  - passes when every checkout disables persistence;
  - fails when a checkout omits `persist-credentials: false` (reproduces the
    pre-fix state);
  - fails when only one of two checkouts disables it;
  - honours a documented `BP-PERSIST-CREDS` ignore;
  - fails when no checkout step is present;
  - errors on a missing workflow file / unknown flag;
  - asserts the real repo `security.yml` passes the rule.
- All new BATS tests pass; `shellcheck` and `actionlint` are clean.

### Note on the local quality gate

`./quality.sh` reports one unrelated failure —
`tests/scripts/cargo_metadata.bats::cargo metadata exposes the repository
field` — because this local checkout has no sibling `NEAT-AI-core` clone at
`../../NEAT-AI-core/neat-core`, so `cargo metadata` cannot resolve the
`neat-core` path dependency. It is environmental (CI checks the sibling out, as
this very workflow does) and cannot be affected by a workflow-YAML / shell /
docs change. Every workflow-hygiene check and the new suite pass.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -21,6 +21,10 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd  # v5
+        with:
+          # No step in this job pushes back or fetches a private submodule, so
+          # the GITHUB_TOKEN need not be written to .git/config (Issue #388).
+          persist-credentials: false
 
       # Path strategy for the NEAT-AI-core sibling (Issue #18):
       #   * `actions/checkout` refuses any `path:` that resolves outside
@@ -36,6 +40,9 @@ jobs:
           repository: stSoftwareAU/NEAT-AI-core
           ref: Develop
           path: NEAT-AI-core
+          # Public sibling clone, read-only — no credential need be persisted
+          # to disk for the rest of the job (Issue #388).
+          persist-credentials: false
 
       - name: Link NEAT-AI-core sibling path expected by Cargo
         run: |
--- a/README.md
+++ b/README.md
@@ -770,6 +770,30 @@ lives in the called workflow's own job (`security.yml`).
 The rule is validated by `scripts/check-workflow-timeouts.sh` (wired into
 `quality.sh`) and covered end-to-end by `tests/scripts/workflow_timeouts.bats`.
 
+### Checkout credential persistence (Issue #388)
+
+Every `actions/checkout` step in the reusable `security.yml` sets
+`persist-credentials: false`. By default checkout writes the workflow's
+`GITHUB_TOKEN` into `.git/config` as an auth header, where any later step in the
+same job — including a compromised dependency or an injected script — can read
+it and act as the token. The `security` job only reads the checked-out code and
+runs `cargo audit` / dependency-review; it never pushes back and never fetches a
+private submodule, so keeping the credential on disk is pure blast radius.
+
+```mermaid
+flowchart LR
+    A[checkout default] -->|token written to .git/config| B[later step reads it]
+    B --> C[acts as GITHUB_TOKEN]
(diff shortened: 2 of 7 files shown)


nleck (approved):

nleck:
Please fix markdown

stservice:
Fixed the MD040 Markdown Lint failure: added a `text` language to the tool-output fenced block at `docs/archive/pr-summaries/pr-summary-388.md:43` (commit bd83ea8). This addresses the 'Please fix markdown' feedback. Tracked in stSoftwareAU/VibeCoding#3534.

Issue #252, opened 2026-06-23
ci(scorer): gate that fails on an unhandled breaking neat-core bump (keep path dep)


stservice:
Part of #248 (release-process redesign epic). **scorer repo.**

## Problem
scorer consumes neat-core via an unpinned `path` dependency that always tracks head:

```toml
# rust_scorer/Cargo.toml
neat-core = { path = "../../NEAT-AI-core/neat-core" }
```

(CI clones NEAT-AI-core into the workspace and symlinks it — see the Cargo.toml comment and `.github/workflows/*`.) The user explicitly chose to **keep the path dep** (Round 2) rather than move to a versioned dependency. So the safeguard must be a **CI gate that fails when neat-core has a breaking bump scorer has not yet handled**, forcing a deliberate upgrade instead of tracking head blindly. This is what would have caught neat-core #177 before the build broke.

## Task
Add a scorer CI gate (a new step in `ci.yml` or a dedicated workflow) that fails on an unhandled breaking neat-core bump.

**Recommended mechanism — version-baseline check** (pairs with the neat-core semver sub-issue):
- Record a "last-handled" neat-core version in scorer (e.g. a checked-in `neat-core.expected-version` file, or a field in scorer config).
- In CI, read neat-core's actual version from the cloned `../../NEAT-AI-core/Cargo.toml` (`[workspace.package] version`).
- **Fail** if neat-core's major-equivalent component (pre-1.0: the minor) is **greater** than the recorded baseline — a breaking bump scorer hasn't acknowledged. **Pass** when they match (patch-level drift allowed per policy).
- "Handling" = a deliberate scorer PR that makes the corresponding code change **and** bumps the recorded baseline to the new neat-core version.

**Alternative mechanism (fallback if neat-core semver isn't ready):** a type/API-surface diff — e.g. `cargo public-api`, or a checked-in snapshot of the consumed neat-core types — that fails CI when the consumed surface changes unexpectedly.

## Acceptance criteria
- [ ] Path dep `neat-core = { path = … }` is **unchanged** (kept, per Round 2 decision).
- [ ] CI **fails** when neat-core presents a breaking bump above scorer's recorded baseline; the failure message points at the required deliberate-upgrade step.
- [ ] CI **passes** after scorer records handling (baseline bumped + code updated).
[...]

PR #477, opened 2026-07-29
fix: reject record-misaligned training files up front (Issue #476)

### LINKED ISSUE #476: bug: continuous .bin stream splices records across file boundaries when a training file is not record-aligned
## Symptom

Two hosts scoring the **same creature** against the **same training corpus**
disagree by a small, systematic, one-direction offset (~+0.0002). Several hosts
in a scoring fleet also logged, at the very end of a run:

```
Error: Trailing 3696 bytes (incomplete record) after reading all training files
```

and then silently fell back to WASM scoring, so the two engines scored
*different record sets* and produced different scores for identical inputs.

## Root cause

Forward-only scoring streams **all** `.bin` files as one continuous byte stream
(`neat_core::training_bin_stream::for_each_read_chunk` feeding
`rust_scorer/src/stream_io.rs::run_io_loop`). The `pending` buffer carries a
short tail from file N straight into file N+1. So if any single `.bin` file's
length is not a whole multiple of
`record_bytes = (num_inputs + num_outputs) * 4`, the native scorer **splices**
a bogus record across the file boundary out of the tail of one file and the
head of the next, and every record after that boundary is shifted.

Two things make this hard to see:

1. The only complaint is the end-of-stream `Trailing N bytes (incomplete
   record) after reading all training files` message, which arrives after the
   whole corpus has been read and **names no file**.
2. If two files' misalignments cancel out (say one file is `record_bytes + 48`
   and another `2 * record_bytes - 48`), the concatenated total *is* a whole
   number of records, so the trailing-bytes check reports nothing at all — even
   though every record after the first boundary is spliced.

The WASM/JS scorer frames records **per file** and asserts on misalignment, so
it never splices. That asymmetry is what turns a misaligned corpus into a
reproducible native-vs-WASM score divergence.

## Proposed fix

Pre-flight guard in `rust_scorer` (new `corpus_guard` module), called
immediately after every `find_bin_files(...)` site and before any streaming:
for each file, `fs::metadata(path).len() % record_bytes` must be `0`. On
failure, fail loudly naming the offending file, its size and the remainder
bytes. Reject `record_bytes == 0` rather than dividing by zero, and propagate
I/O errors rather than skipping unreadable files. Keep the existing
trailing-bytes checks as a backstop.

No `neat-core` change required — the guard lives entirely in this repo.

stservice:
Closes #476. Root cause for stSoftwareAU/GRQ#3753.

## The splice mechanism

Forward-only scoring streams **all** `.bin` training files as one continuous
byte stream: `neat_core::training_bin_stream::for_each_read_chunk` feeding
`rust_scorer/src/stream_io.rs::run_io_loop`. The `pending` buffer carries a
short tail from file N straight into file N+1.

So if any single `.bin` file's length is not a whole multiple of
`record_bytes = (num_inputs + num_outputs) * 4`, the native scorer **splices**
a bogus record across the file boundary — its head is the tail of file N, its
tail is the head of file N+1 — and every record after that boundary is shifted
by the remainder. The scoring run completes and emits a perfectly well-formed
JSON result; it is just computed over a different record set.

Two things kept this invisible:

1. The only existing complaint is the end-of-stream message `Trailing N bytes
   (incomplete record) after reading all training files`. It arrives after the
   entire corpus has been read and **names no file**, so there is nothing to
   act on.
2. When two files' misalignments cancel out — one file `record_bytes + 48`,
   another `2 * record_bytes - 48` — the concatenated total *is* a whole number
   of records, so the trailing-bytes check reports **nothing at all**, even
   though every record after the first boundary is spliced.

The WASM/JS scorer in `@stsoftware/neat-ai`
(`src/creature/CreatureActivation.ts::evaluateDir`) frames records **per file**
and asserts on misalignment, so it never splices. That asymmetry is what turns
a misaligned corpus into a reproducible native-vs-WASM score divergence.

## Fleet evidence

One fleet host scored the same creature ~**+0.0002** higher than six other
hosts on the same training corpus. Four other hosts logged the identical
message and quietly changed engines mid-fleet:

```
Rust scorer call failed (exit 1); stderr: Error: Trailing 3696 bytes
(incomplete record) after reading all training files; falling back to WASM
scoring.
```

Hosts that fell back scored the per-file-framed record set; hosts that stayed
native scored the spliced one — hence a small, systematic, one-direction offset
rather than random noise.

## The fix

New `rust_scorer/src/corpus_guard.rs`:

```rust
pub fn assert_records_aligned(bin_files: &[PathBuf], record_bytes: usize) -> Result<(), String>
```

For every file, `fs::metadata(path).len() % record_bytes` must be `0`. It is
called immediately after every `find_bin_files(...)` site — `src/main.rs` and
both `src/multi_score.rs` sites (CPU and GPU directory paths) — **before any
streaming begins**, using `config.bytes_per_record()`.

On failure the run aborts with the offending file named:

```
Training file /data/train/b.bin has size 30 bytes, which is not a whole
multiple of the 8-byte record size (6 bytes past the last whole record);
records would be spliced across file boundaries
```

When several files are misaligned the message names the first and adds
`(N other training files are also misaligned)`.

Also:

* `record_bytes == 0` is rejected with a clear message rather than dividing by
  zero.
* An unreadable or missing file surfaces as an error — never a silent skip.
* The existing "Trailing N bytes" checks are **untouched** and stay as a
  backstop.
* **No `neat-core` change**, so this needs no cross-repo dependency bump.

## Tests

TDD, calling the real function:

* `corpus_guard::tests::aligned_files_pass`
* `corpus_guard::tests::single_misaligned_file_is_rejected_and_named`
* `corpus_guard::tests::offsetting_misalignments_are_still_rejected` — the case
  the trailing-bytes check misses entirely
* `corpus_guard::tests::several_misaligned_files_are_counted`
* `corpus_guard::tests::zero_record_bytes_is_rejected`
* `corpus_guard::tests::missing_file_surfaces_an_error`
* `corpus_guard::tests::empty_corpus_is_aligned`

Plus an end-to-end CLI test,
`scorer_smoke::scorer_binary_rejects_misaligned_training_files`, which builds a
corpus whose two misalignments cancel out and asserts the binary exits non-zero
naming the file.

## Quality gate

`./quality.sh` passes except for one **pre-existing, unrelated** failure: the
neat-core breaking-bump gate (`scripts/check-neat-core-version.sh`) reports the
sibling neat-core at `0.5.0` against the `0.2.5` baseline in
`neat-core.expected-version`. That failure reproduces on a clean `Develop`
checkout with these changes stashed. Every step after it was run individually
and is clean: shellcheck, all validator scripts, bats (434 tests), cargo-deny,
`cargo fmt --check`, clippy (`-D warnings -D clippy::filter_next -D
clippy::collapsible_if`), `cargo check`, debug build, `cargo test --workspace
--all-features` (23 suites, 0 failures), rustdoc with `RUSTDOCFLAGS=-D
warnings`, and the release build.

`Cargo.lock` was already modified in the working tree before this change and
has deliberately been left unstaged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,26 @@ section to the released version with its date.
 
 ## [Unreleased]
 
+### Fixed
+
+- **Misaligned training files no longer splice records silently (Issue #476).**
+  The native scorer streams every `.bin` file as one
+  continuous byte stream, so the `pending` buffer carries a short tail from file
+  N straight into file N+1. Any single file whose length is not a whole multiple
+  of `record_bytes` therefore produced a bogus spliced record at the boundary and
+  shifted every record after it — the run only complained at the very end, with
+  `Trailing N bytes (incomplete record) after reading all training files`, naming
+  no file, and said nothing at all when two misalignments cancelled out across the
+  corpus. The WASM scorer frames records per file and asserts, so hosts that fell
+  back to WASM scored a different record set from hosts on native — a small,
+  systematic, one-direction score offset across the fleet. New
+  `rust_scorer/src/corpus_guard.rs::assert_records_aligned` runs immediately after
+  every `find_bin_files(...)` site, before any streaming, and fails loudly naming
+  the offending file, its size and the remainder bytes. Zero `record_bytes` is
+  rejected rather than dividing by zero, and an unreadable file is an error rather
+  than a silent skip. The end-of-stream trailing-bytes checks stay as a backstop.
+  No `neat-core` change.
+
 ### Removed
 
 - **Scorer-local dead-code audit (Issue #470).** Re-ran the superseded
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,17 @@ section to the released version with its date.
 
 ## [Unreleased]
 
+### Changed
+
+- **neat-core baseline acknowledged at 0.5.0 (Issue #252 gate).** neat-core has
+  presented three pre-1.0 breaking bumps since the recorded 0.2.5 baseline, each
+  removing API `rust_scorer` does not consume: the deprecated `score_records` /
+  `score_records_parallel` wrappers and `RecordBatch::PerRecord` (0.3.0),
+  `PredictiveCodingEngine` (0.4.0), and the `wasm_dataset` training-data offload
+  (0.5.0). scorer's migration to the flat scoring entry points had already
+  landed, so no scorer code change was outstanding; verified by a clean build
+  and the full test suite against neat-core 0.5.0.
+
 ### Fixed
 
 - **Misaligned training files no longer splice records silently (Issue #476).**
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -750,7 +750,7 @@ dependencies = [
 
 [[package]]
 name = "neat-core"
-version = "0.2.25"
+version = "0.5.0"
 dependencies = [
  "serde",
  "serde_json",
@@ -1076,7 +1076,7 @@ checksum = "19b30a45b0cd0bcca8037f3d0dc3421eaf95327a17cad11964fb8179b4fc4832"
 
 [[package]]
 name = "rust_scorer"
-version = "1.1.35"
+version = "1.1.37"
 dependencies = [
  "bytemuck",
  "clap",
--- a/neat-core.expected-version
+++ b/neat-core.expected-version
@@ -11,4 +11,18 @@
 # rust_scorer handling landed in milestone #257; patch drift within the 0.2 line
 # is non-breaking. Verified by a clean `cargo build -p rust_scorer` and the full
 # scorer test suite passing against the sibling neat-core.
-0.2.5
+#
+# 0.2.5 -> 0.5.0: acknowledge the three breaking bumps neat-core has presented
+# since 0.2.5. Each removes API that rust_scorer does not consume:
+#   * 0.3.0 — neat-core #412 deleted the deprecated `score_records` /
+#     `score_records_parallel` wrappers and `RecordBatch::PerRecord`. scorer had
+#     already migrated to the flat scoring entry points (neat-core #411), so no
+#     scorer change was outstanding.
+#   * 0.4.0 — neat-core #420 removed `PredictiveCodingEngine`
+#     (pc_inference.rs / pc_learning.rs); scorer never referenced it.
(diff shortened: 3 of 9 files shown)


stservice:
## CI fix: Project Validation — neat-core breaking-bump gate

**Root cause.** The failing step was `scripts/check-neat-core-version.sh`, not
anything in this PR's diff. The sibling neat-core has advanced to **0.5.0**
while `neat-core.expected-version` still recorded **0.2.5**, so the Issue #252
breaking-bump gate failed — and would fail on every open PR until acknowledged.

**Investigation.** All three pre-1.0 breaking bumps since 0.2.5 remove API that
`rust_scorer` does not consume:

| Core version | Change | scorer impact |
| --- | --- | --- |
| 0.3.0 | neat-core #412 — deleted deprecated `score_records` / `score_records_parallel` and `RecordBatch::PerRecord` | none; scorer already migrated to the flat entry points (neat-core #411) |
| 0.4.0 | neat-core #420 — removed `PredictiveCodingEngine` | never referenced |
| 0.5.0 | neat-core #421 — removed the `wasm_dataset` offload | never referenced |

Step 1 of the gate's remediation ("update rust_scorer for the breaking change")
was therefore already satisfied, so this commit performs step 2 — bumping the
recorded baseline, with the reasoning written into the file itself rather than a
bare version bump.

**Verification against the sibling neat-core at 0.5.0** (`origin/Develop`, the
same ref CI clones):

- `cargo build -p rust_scorer` — clean.
- `cargo test -p rust_scorer` — 335 tests, 0 failures.
- `./quality.sh` — ✅ all quality checks passed.
- `./scripts/check-neat-core-version.sh` — `OK neat-core 0.5.0 matches handled baseline 0.5.0`.
- Repo-wide grep for `score_records`, `RecordBatch::PerRecord`,
  `PredictiveCoding`, `wasm_dataset` — no hits in `rust_scorer/` or `tests/`.

**Files changed:** `neat-core.expected-version`, `CHANGELOG.md`. No production
code was touched, and the gate itself was not weakened or skipped — it now
passes because the condition it guards is genuinely satisfied.

**Reviewer note.** The gate is designed to force a *deliberate* acknowledgement.
That deliberation is recorded above and in the version file's comment block; if
you would rather this land as its own standalone upgrade PR, drop commit
`3eb4dd2` from this branch and cherry-pick it — this PR's own fix (Issue #476)
is independent of it.

Issue #474, opened 2026-07-29
dead-code: 27 pub items are crate-internal and should be pub(crate)


VibeCoderST:
Found by the dead-code audit in [stSoftwareAU/NEAT-AI-core#413](https://github.com/stSoftwareAU/NEAT-AI-core/issues/413).

## Finding

`rust_scorer` is not published, so `pub` on an item with no consumer outside the crate widens the API surface for nothing — and, worse, it suppresses the `dead_code` lint that would otherwise catch the item going stale. **27 items are only ever used inside the crate** and should be `pub(crate)`.

| path:line | symbol |
| --- | --- |
| `src/scoring.rs:13` | `SEMANTIC_MAJOR_VERSION` |
| `src/read_tuning.rs:18` | `MAX_READ_BYTES` |
| `src/read_tuning.rs:21` | `default_training_read_bytes` |
| `src/shallow_fixture.rs:27` | `ENCELADUS_SQUASHES` |
| `src/shallow_fixture.rs:61` | `plan_synapses` |
| `src/prod_fixture.rs:47,49,51,53,55,57` | `MIN_INPUTS`, `EXPECTED_OUTPUTS`, `MIN_NEURONS`, `MAX_NEURONS`, `MIN_SYNAPSES`, `MAX_SYNAPSES` |
| `src/prod_fixture.rs:99` | `parse_production_creature` |
| `src/prod_fixture.rs:110` | `check_production_topology` |
| `src/gpu/forward_mse_batched.rs:74` | `MAX_NEURONS_ABSOLUTE` |
| `src/gpu/forward_mse_batched.rs:80` | `DEFAULT_SCRATCH_BUDGET_BYTES` |
| `src/gpu/forward_mse_batched.rs:125,152,161` | `NeuronGpu`, `CreatureMetaGpu`, `BatchedNetworkData` |
| `src/gpu/forward_mse_batched.rs:235` | `build_batched_network_data` |
| `src/gpu/forward_mse_batched.rs:436` | `BatchedRunner::from_data` |
| `src/gpu/forward_mse_batched.rs:1012` | `DirectoryGpuRunners` |
| `src/gpu/forward_mse_batched.rs:1102` | `DirectoryGpuRunners::kernel_label` |
| `src/gpu/forward_mse_batched.rs:1122` | `scratch_workgroups_x_for` |
| `src/gpu/mod.rs:151` | `GpuBackendLabel::from_wgpu` |
[...]

PR #479, opened 2026-07-29
dead-code: main.rs re-declares the module tree, forcing 16 #[allow(dead_code)] suppressions (Issue #475)

### LINKED ISSUE #475: dead-code: main.rs re-declares the module tree, forcing 16 #[allow(dead_code)] suppressions
Found by the dead-code audit in [stSoftwareAU/NEAT-AI-core#413](https://github.com/stSoftwareAU/NEAT-AI-core/issues/413). This is the **root cause** behind the crate's `#[allow(dead_code)]` sites and behind a dozen `pub` markers that cannot be tightened (see #474).

## Finding

`src/main.rs:13-21` re-declares its own `mod` tree instead of consuming the crate through `use rust_scorer::…`. The bin target therefore compiles a **second, independent copy** of every module. Anything used only from the lib side is dead in the bin copy, so the crate has to blanket-suppress it.

`rust_scorer/Cargo.toml` already declares both targets:

```toml
[lib]
name = "rust_scorer"
path = "src/lib.rs"

[[bin]]
name = "rust_scorer"
path = "src/main.rs"
```

## Evidence

Every one of the 16 `#[allow(dead_code)]` sites in `src/` was stripped in a throwaway copy, then:

```
cargo check --workspace --all-targets
→ 19 dead_code errors, ALL in the `bin "rust_scorer"` target
  (one per stripped site, plus 2 cascades: InvalidCostName, ResolveBackendError)
```

So **none of the attributes is stale** — each is load-bearing today. They are all load-bearing *because of the duplicated module tree*, not because the code is genuinely unused. Beyond compile cost, the duplication means the bin and the lib can drift: a change to the lib copy is not necessarily what the shipped binary runs.

## What a fix looks like

Replace the `mod` declarations in `main.rs` with `use rust_scorer::…` so the bin links the lib rather than recompiling it. Then, in the same pass:

- delete the 16 `#[allow(dead_code)]` attributes that exist only to silence the bin copy, and
- downgrade the twelve `pub` items listed under "*`pub` currently acts as `dead_code` suppression*" in #474 to `pub(crate)`.

That re-arms `dead_code` across the crate, so genuinely dead code starts failing the build instead of accumulating behind suppressions.

## Also worth correcting while in the area

Several `neat-core` paths are referenced **only inside doc comments** in `rust_scorer` and no longer correspond to a real call: `neat_core::batch_scoring`, `neat_core::synapse_type`, `SynapseData`, `reset_state`, `activate`, `apply_squash`, `apply_limit_range`. `rust_scorer` actually consumes ~25 `neat-core` public items (`creature::*`, the `loss::*_sum_batch_packed` family, `CompiledNetwork`/`activate_into`, `training_data::*`, `training_bin_stream::for_each_read_chunk`, `SquashType`).

Filed as a report-only finding; the change rides its own PR against this repo.

stservice:
# dead-code: the binary links the library instead of recompiling it

## Summary

`rust_scorer/src/main.rs` declared its own `mod` tree, so the `rust_scorer`
**bin** target compiled a **second, independent copy** of every module. Anything
used only from the lib side was dead in the bin copy, which is why the crate
carried 17 `#[allow(dead_code)]` attributes and a dozen `pub` markers that could
not be tightened. Worse, the two copies could drift: a change to the library was
not necessarily what the shipped binary ran.

The CLI logic moved verbatim to `rust_scorer/src/cli.rs` (a module of the
library) and `main.rs` is now a nine-line shim over `rust_scorer::cli::main`. In
the same pass all 17 suppressions were deleted and ten `pub` items that existed
only to dodge `dead_code` were downgraded to `pub(crate)`, so the lint is armed
across the whole crate. No behaviour, JSON, or CLI-contract change.

Closes #475.

```mermaid
flowchart LR
    subgraph Before["Before — two copies"]
        M1["main.rs<br/>mod cost; mod gpu; …"] --> C1["module copy A<br/>(bin target)"]
        L1["lib.rs<br/>pub mod cost; …"] --> C2["module copy B<br/>(lib target)"]
        C2 --> T1["benches + integration tests"]
        C1 -. "unused in bin ⇒<br/>17 #[allow(dead_code)]" .-> C1
    end
    subgraph After["After — one copy"]
        M2["main.rs (shim)<br/>rust_scorer::cli::main()"] --> L2["lib.rs<br/>pub mod cli; pub mod cost; …"]
        L2 --> C3["one module tree"]
        C3 --> T2["benches + integration tests"]
    end
```

## Changes

- **`rust_scorer/src/cli.rs`** (renamed from `src/main.rs`) — the `mod …;` block
  is replaced by `use crate::{cost, gpu, multi_score, scoring, stream_score};`
  and `fn main` becomes the documented `pub fn main`. Everything else, including
  all 24 CLI unit tests, is unchanged; they now run in the lib test target.
- **`rust_scorer/src/main.rs`** — thin shim calling `rust_scorer::cli::main()`.
- **`rust_scorer/src/lib.rs`** — declares `pub mod cli;`; stale module-doc about
  the binary keeping its own tree corrected.
- **17 `#[allow(dead_code)]` attributes deleted** across `multi_score.rs` (8),
  `gpu/mod.rs` (3), `cost.rs` (2), `sampling.rs` (2), `stream_score.rs` (1) and
  `gpu/forward_mse_batched.rs` (1), along with the comments explaining them.
- **Ten `pub` → `pub(crate)`**: `MAX_SHALLOW_NON_INPUT_NEURONS`,
  `directory_gpu_topology`, `directory_pool_is_shallow`,
  `auto_should_use_gpu_directory`, `auto_topology_fallback_note`,
  `auto_cost_fallback_note`, `resolve_mode`, `DirectoryGpuProbe`,
  `gpu_directory_probe_for_dir`, `RecordSampler::keep_next`.
- Doc comments naming `main.rs` as the CLI call site now name `cli.rs`.
- `CHANGELOG.md` — entry under Unreleased/Changed.

### Three items on #474's list deliberately stayed `pub` (documented inline)

A bin target that links a lib is a **separate crate**, so `pub(crate)` items are
invisible to it — which is exactly why the CLI moved *into* the library rather
than just swapping `mod` for `use`. That reaches all but three:

| Item | Why it stays `pub` |
| --- | --- |
| `ScoringPath` | `SingleCreature` is constructed only by unit tests (Issue #81 settled that path as CPU-only). `dead_code` ignores `cfg(test)` construction, so `pub(crate)` would need a **new** `#[allow(dead_code)]` — the suppression this issue removes. |
| `parse_sample_rate` | Has its own doctest; doctests compile as external crates, so `pub(crate)` breaks `cargo test --doc` (#474's own doctest-only carve-out). |
| `RecordSampler` | Same — `RecordSampler::filter_in_place` is doctested externally. |

`auto_should_use_gpu` also stays `pub`: it is linked from public module docs, so
downgrading it trips `rustdoc::private_intra_doc_links` under `-D warnings`.

### "Also worth correcting" — verified, no change needed

The issue flagged seven `neat-core` paths referenced only in doc comments as
possibly stale. All seven were checked against the sibling `NEAT-AI-core` clone
and **every one still exists**: `batch_scoring` (`src/batch_scoring.rs`),
`synapse_type` (`src/synapse_type.rs`), `SynapseData`, `apply_squash`
(`src/squash.rs`), `apply_limit_range` (`src/range.rs`), `reset_state` and
`activate` (`src/network.rs:215`/`373`). They are accurate cross-references
describing what the WGSL kernels mirror on the CPU side, not dangling links, so
they were left alone rather than churned.

## Evidence

Backend/CLI change — no web interface, so no screenshot. Evidence is the build
gate plus the new parity tests.

**The suppressions are genuinely gone, not relocated.** `dead_code` is now armed
crate-wide and the full gate is clean:

```console
$ grep -rnE '^[[:space:]]*#\[allow\(dead_code\)\]' rust_scorer/src/
(no matches — only two prose mentions remain, both explaining why an
 attribute is *not* there)

$ ./quality.sh < /dev/null
...
✅ All quality checks passed!
```

That gate runs `RUSTFLAGS="-D warnings"` over `cargo clippy --all-targets`,
`cargo build`, `cargo test --workspace`, `RUSTDOCFLAGS="-D warnings" cargo doc`
and a release build — so any item that lost its last caller now fails the build
instead of hiding behind an attribute. Re-arming immediately caught one real
finding (`ScoringPath::SingleCreature`, table above).

**No tests lost in the move.** The 24 CLI unit tests that lived in `main.rs`
moved with it and now run in the lib target:

```text
Running unittests src/lib.rs   ... 182 passed; 0 failed
Running unittests src/main.rs  ...   0 passed; 0 failed   (shim has no tests)
Doc-tests rust_scorer          ...  24 passed; 0 failed
```

## Test Plan

New — `rust_scorer/tests/bin_lib_single_source.rs` (2 tests). These assert the
invariant behaviourally rather than by inspecting source: the JSON the **binary**
prints must equal, value for value, what the **library** entry point returns for
the same inputs. A future re-duplication that let the two drift shows up here as
a numeric mismatch.

- `binary_directory_scores_match_library_entry_point` — runs the compiled binary
  (`--gpu off`, directory mode) over a two-creature fixture and compares
  `error`, `score`, `recordCount` and `complexityPenalty` per creature against
  `rust_scorer::multi_score::score_from_creature_dir`.
- `binary_and_library_agree_for_non_default_cost` — same under `--cost MAE`, so
  the shared cost dispatch (not just the MSE default) is proven to come from one
  implementation, and checks the binary echoes `costName: "MAE"`.

Existing — unchanged and passing: 182 lib unit tests (including the 24 CLI tests
relocated from `main.rs`), 24 doctests, and every integration suite
(`scorer_smoke`, `directory_mode_tdd`, `cost_parity`, the GPU parity suites,
`single_pass_assertion`, `compile_once_assertion`, `sample_rate_*`).

## Security Self-Check

- Input validation: unchanged — the same clap parsing, `parse_sample_rate`
  range check and `assert_records_aligned` corpus guard run on the same inputs.
- Secrets: none staged; no hidden files touched.
- Injection surface: none added; no new SQL, shell, filesystem or HTTP calls.
- Error handling: unchanged — `run` still returns `Result<_, String>` and
  `cli::main` still exits `1` after an `Error: …` line on stderr.
- Dependencies: none added or changed (`Cargo.lock` diff is only the pre-existing
  `rust_scorer` version-bump drift, 1.1.37 → 1.1.38, matching `Cargo.toml`).

---

🤖 Processed by: stservice
🏷️ Run id: `vibe-ms5ya3z5-6e9fa2`


stservice:

```
Error: cargo not available

```

</details>

stservice:
## Markdown Lint fix

**Root cause:** `docs/archive/pr-summaries/pr-summary-475.md` had two fenced code
blocks with no language, tripping `MD040/fenced-code-language` at lines 89 and 108.
That was the whole of the `Process completed with exit code 1` failure.

**Fix:** tagged the blocks — the shell transcript as ` ```console `, the test-output
block as ` ```text `. No rule suppressions, no config changes.

**Verified locally:** `markdownlint-cli2` → `Summary: 0 error(s)` across all 155 files.

Committed as `docs: add languages to fenced code blocks (PR #479)`.

Issue #435, opened 2026-07-22
Should not need manual approval to run actions


nleck:
The other repos like NEAT-AI don't 

![image](https://github.com/user-attachments/assets/8abc7783-25d4-4ba4-b56f-97e1e741c0a8)

VibeCoderST:
## Needs a human — this is a repository Actions **setting**, not a code change

**Root cause.** The "⚠️ N checks awaiting approval — This workflow requires approval from a maintainer / **Approve and run**" banner in the screenshot is GitHub's *fork / first-time-contributor workflow-approval gate*. It is controlled entirely by a **repository (or organisation) Actions setting**, not by anything in the committed workflow YAML. No change to `.github/workflows/*.yml` can remove it.

**What I verified this run**
- `VibeCoderST` already has **`write`** access on both `NEAT-AI-scorer` and `NEAT-AI`, so a missing collaborator grant is *not* the difference.
- The repo is not a fork (`fork:false`) and the bot has `admin:false` / `maintain:false`.
- Every Actions-policy endpoint returns **403** for the worker account, including the exact one that governs this gate:
  - `GET repos/stSoftwareAU/NEAT-AI-scorer/actions/permissions/fork-pr-contributor-approval` → `403 "You must have repository read permissions or have the repository Actions policies fine-grained permission."`

So the worker account **cannot read or change** this setting — it needs admin.

**What a human needs to do** (someone with **admin** on the repo/org):

1. Open **`Settings → Actions → General`** on `stSoftwareAU/NEAT-AI-scorer`:
   <https://github.com/stSoftwareAU/NEAT-AI-scorer/settings/actions>
2. In the **"Fork pull request workflows from outside collaborators"** section, change **"Require approval for …"** so it matches `NEAT-AI` (which does not gate the bot's PRs — e.g. set it to the less strict option there).
3. Compare the same page on the working repo for reference:
   <https://github.com/stSoftwareAU/NEAT-AI/settings/actions>

Equivalent via the API (admin token required):
```bash
# Inspect both repos to see the difference:
gh api repos/stSoftwareAU/NEAT-AI/actions/permissions/fork-pr-contributor-approval
gh api repos/stSoftwareAU/NEAT-AI-scorer/actions/permissions/fork-pr-contributor-approval
# Then align NEAT-AI-scorer to match NEAT-AI (set approval_policy accordingly).
```

[...]

Issue #100, opened 2026-05-18
[security] supply-chain: GitHub Actions pinned to mutable refs across CI workflows


VibeCoderST:
## Summary

GitHub Actions across every CI workflow except `markdown-lint.yml` are pinned to mutable refs (version tags like `@v5`, `@v4`, `@v8`, `@stable`, `@2.0.0`) rather than to 40-character commit SHAs. A compromised maintainer of any of these third-party actions — or a re-tag of a mutable ref — would silently re-execute under the privileges of these workflows.

The repository already documents a SHA-pinning policy in `.github/workflows/markdown-lint.yml` (lines 12–15: "actions/checkout — SHA-pinned to v5 (Node 24). actions/setup-node — SHA-pinned to v4 (Node 24)") and applies it correctly on lines 37 and 40. The standard exists but is inconsistently applied across the remaining workflows.

## Class

`supply-chain:unpinned-actions`

## Severity

**Medium** — `AV:N/AC:H/PR:N/UI:N/S:C/C:H/I:H/A:H`. Exploit requires the attacker to compromise a third-party action's release pipeline or re-push a tag, but on success they execute under workflows with `contents: write` (auto-format, upgrade-dependencies) and `pull-requests: write` (upgrade-dependencies), including access to `GITHUB_TOKEN`.

## Confidence

**high** — every offending line is a direct citation. Confidence lift: an observed historical tag re-push for one of the pinned actions would raise impact certainty.

## Ease of exploit

**hard** — requires compromise of an upstream action repository or maintainer credentials. Several of the actions involved (e.g. `actions/*`) are first-party GitHub, but third-party ones (`dtolnay/rust-toolchain`, `ludeeus/action-shellcheck`, `peter-evans/create-pull-request`, `rustsec/audit-check`) materially widen the attack surface. `dtolnay/rust-toolchain@stable` is especially exposed because `@stable` is a moving ref tracking the upstream's chosen "stable" alias.

## Attacker model

External attacker who compromises a third-party action publisher (or who re-pushes a release tag) and waits for any workflow run on the `Develop` branch.

## Trigger

Any CI workflow run on `Develop` (push, PR, scheduled). The auto-format workflow and the weekly `Upgrade Cargo Dependencies` workflow both have `contents: write`, so a successful compromise can push to the repository.

## Why it is a bug

Sample of unpinned references — every one is a tag, not a 40-character SHA:

```
[...]

PR #501, opened 2026-08-01
Scope the bot-push credential: replace org-level ACTIONS_PUSH PAT with a short-lived repo-scoped token (Issue #498)

### LINKED ISSUE #498: Scope the bot-push credential: replace org-level ACTIONS_PUSH PAT with a short-lived repo-scoped token

stservice:
# Scope the bot-push credential to a short-lived repo-scoped token

## Summary

`auto-format.yml` and `version-increment.yml` pushed back to the PR head
branch with the **organisation-level** `ACTIONS_PUSH` PAT. That PAT is
long-lived and org-scoped, so anything that reached it stepped up from
single-repo write access to write access on every repository in the
organisation — the residual risk left open by the Issue #497 hardening.

Both jobs now mint their push credential per run with
`actions/create-github-app-token` (SHA-pinned `bcd2ba4…`, `# v3`): a GitHub App
installation token narrowed to `permission-contents: write` and scoped to
**this repository only**, expiring within the hour and revoked by the action's
post step. Pushes remain attributed to a trusted non-`GITHUB_TOKEN` identity,
so the Issue #435 behaviour (checks run on the resulting `synchronize` event
without an "Approve and run" gate) is preserved.

Creating the App and storing its secrets needs organisation-admin access, so
the change is written to degrade safely: a job-level `PUSH_APP_CONFIGURED`
flag (the `secrets` context is unavailable in a step-level `if:`) skips the
mint step until both secrets exist, and `GH_PAT` falls back to
`secrets.ACTIONS_PUSH || secrets.GITHUB_TOKEN`. The workflows therefore behave
exactly as before until an admin completes the handover documented in the
README.

Closes #498.

## Evidence

This is a CI/workflow change with no web interface to screenshot. Verification
is the BATS suite plus the repository's own workflow checkers.

```mermaid
flowchart LR
    A[PR push] --> B{"App secrets set?"}
    B -->|yes| C["Mint installation token<br/>contents: write, this repo, &lt;1h"]
    B -->|no| D["Fallback:<br/>ACTIONS_PUSH PAT → GITHUB_TOKEN"]
    C --> E[Commit and push step]
    D --> E
    E --> F["synchronize event —<br/>PR checks run unblocked"]
    C -.->|post step| G[Token revoked]
```

`scripts/check-bot-push-token.sh` was rewritten from a single `grep` for
`secrets.ACTIONS_PUSH || secrets.GITHUB_TOKEN` into an indentation-aware
scanner (same approach as `check-push-step-hardening.sh`) enforcing four rules
per guarded workflow:

1. the push credential is minted with `actions/create-github-app-token`,
   SHA-pinned per the supply-chain policy (Issue #100);
2. the mint step requests `permission-contents: write`;
3. the mint step scopes the token with `repositories:` to a single repository
   and sets no `owner:` (which would widen it across the organisation);
4. every `GH_PAT` binding prefers the minted token and keeps the
   `ACTIONS_PUSH || GITHUB_TOKEN` fallback.

Local runs (all green):

- `bats tests/scripts/bot_push_token.bats` — 11/11 pass.
- `actionlint .github/workflows/auto-format.yml .github/workflows/version-increment.yml` — clean.
- `scripts/check-push-step-hardening.sh` — the Issue #497 hardening still
  holds on both push steps after the `GH_PAT` change.
- `scripts/check-workflow-action-versions.sh` — the new action is pinned and
  covered by an explicit `required:3` policy entry (v3 ships a Node 24 runtime
  and takes the non-deprecated `client-id` input).
- `./quality.sh` — full gate passes.

**Supply chain:** `actions/create-github-app-token` v3.2.0 was published
2026-05-12, well outside the 24 h quarantine window for external dependencies,
and is pinned to the tag's commit SHA rather than the mutable tag.

## Human action still required

An organisation admin must create the GitHub App and store two repository
secrets before the scoping takes effect:

| Secret                         | Value                     |
| ------------------------------ | ------------------------- |
| `ACTIONS_PUSH_APP_CLIENT_ID`   | The App's client ID       |
| `ACTIONS_PUSH_APP_PRIVATE_KEY` | The App's PEM private key |

Install the App on this repository only, with the `Contents: Read and write`
repository permission. A fine-grained PAT limited to this single repository,
stored as `ACTIONS_PUSH`, is the lower-effort alternative. Until then the
workflows run on the existing fallback path, unchanged.

## Test Plan

`tests/scripts/bot_push_token.bats` — rewritten around the Issue #498 policy:

- `passes when the push token is a repo-scoped app token with PAT fallback`
- `fails when the push relies on the org-level PAT alone` — **deliberately
  inverted**: this case previously asserted that a bare
  `ACTIONS_PUSH || GITHUB_TOKEN` chain passed. Issue #498 changes that business
  rule, so the test now asserts the opposite. No test was removed or commented
  out.
- `fails when workflow only uses GITHUB_TOKEN` — unchanged (Issue #435).
- `fails when the app token action is not SHA-pinned`
- `fails when the minted token is not narrowed to contents: write`
- `fails when the minted token is not scoped to a single repository`
- `fails when the mint step widens scope with an owner input`
- `fails when the push step ignores the minted token`
- `fails when the ACTIONS_PUSH fallback is dropped`
- `reports a missing workflow file instead of passing silently` — the checker
  fails loud rather than treating an absent workflow as a pass.
- `shipped auto-format and version-increment workflows validate cleanly`

## Files changed

- `.github/workflows/auto-format.yml`, `.github/workflows/version-increment.yml`
  — mint step + `GH_PAT` preference chain + `PUSH_APP_CONFIGURED` job env.
- `scripts/check-bot-push-token.sh` — new four-rule policy checker.
- `scripts/check-workflow-action-versions.sh` — `required:3` policy entry.
- `tests/scripts/bot_push_token.bats` — coverage for every rule.
- `README.md` — new "Bot-push credential" section with the admin handover
  steps; the Issue #497 section now points at the landed fix.

---

🤖 Processed by: stservice
🏷️ Run id: `vibe-msalso3m-52aa84`

### DIFF
--- a/.github/workflows/auto-format.yml
+++ b/.github/workflows/auto-format.yml
@@ -38,6 +38,13 @@ jobs:
     # Skip forks — neither GITHUB_TOKEN nor ACTIONS_PUSH can push to a
     # fork's branch. Formatting on fork PRs is the author's responsibility.
     if: github.event.pull_request.head.repo.full_name == github.repository
+    env:
+      # The `secrets` context is unavailable in a step-level `if:`, so the
+      # push App's configured/not-configured state is hoisted to a job-level
+      # env flag (Issue #498). Until an organisation admin creates the App and
+      # stores its secrets this is 'false': the mint step is skipped and the
+      # push falls back to ACTIONS_PUSH / GITHUB_TOKEN.
+      PUSH_APP_CONFIGURED: ${{ secrets.ACTIONS_PUSH_APP_CLIENT_ID != '' && secrets.ACTIONS_PUSH_APP_PRIVATE_KEY != '' }}
 
     steps:
       - name: Checkout PR branch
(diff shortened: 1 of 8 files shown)


stservice:

```
Error: cargo not available

```

</details>

PR #540, opened 2026-08-06
perf candidate: split directory-mode chunk work into finer sub-ranges so the Rayon tail stops idling threads (Issue #537)

### LINKED ISSUE #537: perf candidate: split directory-mode chunk work into finer sub-ranges so the Rayon tail stops idling threads
## Idea

In directory mode the per-chunk work unit is **one whole chunk for one creature**, so the Rayon pool sees exactly `N` tasks per chunk. Split each creature's chunk into `k` record sub-ranges instead, so the pool sees `k × N` finer tasks and the tail of every chunk stops idling threads.

`rust_scorer/src/multi_score.rs:258` — `workers_per_creature()`:

```rust
if n_creatures >= threads {
    return vec![1; n_creatures];   // N=50, 10 threads → 50 indivisible tasks
}
```

Only when `N < threads` does a creature get split across workers, via `partition_packed_record_ranges()`. **The splitting machinery already exists and is already exercised** — it is simply switched off at the population size production actually runs.

## Why it should help

Two granularity effects, both repeated on **every one of the ~704 chunks** (32 MiB default read buffer ÷ 9,848 B = 3,404 records/chunk):

1. **`ceil()` waste when `N` is not a multiple of the thread count.** 50 tasks on 16 threads is `ceil(50/16) = 4` rounds against an ideal 3.125 → **28 % of the activation budget idle**. On 12 threads: 5 vs 4.17 → 20 %. On the 10-core bench host 50/10 divides exactly, so this effect is *invisible on that machine and real on the rest of the cluster*.
2. **Heterogeneous P/E cores (the bigger one).** An Apple M4 is 4 performance + 6 efficiency cores, and Rayon hands out equal-sized tasks with no idea that E-core threads run ~3× slower. From the baseline table: per chunk the wall is ≈347 ms and one creature-chunk is ≈41 ms of P-core time (≈125 ms on an E core). A single trailing task picked up late by an E-core thread therefore exposes up to ~100 ms of a ~347 ms chunk with P cores idle. Work-stealing cannot fix this below a granularity of one task.

Splitting into `k = 4..8` sub-ranges cuts the exposed straggler to ~1/k and lets fast threads mop up the tail.

## Expected gain

**3–15 %**, host-dependent: near zero on a homogeneous machine where `N` divides the thread count, largest on big Macs (16/24 threads) and on any host with E cores. Worth measuring on more than one cluster machine.

## Change size / risk

**Small.** Relax the `n_creatures >= threads` early return to target `total_workers ≈ k × threads`, and let the existing `worker_offsets` / `work_ranges` / `worker_creature_idx` plumbing do the rest.

Watch:
- **Memory** — one `CompiledNetwork` clone per worker. `k=4` at N=50 is 200 clones (~200 × 300 KB ≈ 60 MB with the #531 `mse_inter` scratch). Keep the `max_worker_count()` RAM ceiling honoured.
- **Numerics** — a creature's total becomes a sum of `k` f64 partials instead of one, so the f64 addition order changes. Fold partials in a fixed worker order so the result is deterministic; expect relative diff < 1e-9, not bit-identity (same tolerance #529 accepted).
- Early-exit (`active[ci]`) must gate all `k` of a creature's workers together.

Cheap first experiment: hard-code `k` behind an env var and sweep `k ∈ {1,2,4,8}` on the production fixture before touching the default.

## Acceptance gate (prototype-first, mandatory)

Nothing merges on a microbenchmark. Prototype, measure on the production fixture, then decide.

| Axis | Production target |
|---|---|
| Creature | Deep production creature (~2,511 inputs / ~1,606 hidden, ~21k synapses) |
| Pool | **N≈50 distinct variants** — randomly perturb weights/biases. Do **not** symlink one `network.json` N times (see scorer#531/#532) |
| Corpus | Full multi-file `.bin` set, ≈22 GiB at 9,848 B/record |
| Mode | Directory scoring, `--gpu off` |
| Method | Interleaved before/after binaries, ≥5 A/B pairs, quiet host, warm cache; compare **medians** |
| Bar | Reproducible **≥1–2 %** median wall-clock gain **and** the change stays minor |
| Parity | Error vectors bit-identical, unless this issue explicitly notes an f32 re-association tolerance |

**Risk/reward rule (parent scorer#536):** a small gain bought with a large or risky change is a losing proposition. If the bar is missed, *or* hitting it needs a big change — post the before/after numbers, label `negative-result`, close as `not planned` per the Performance Task Workflow.

## Measured baseline to beat

From the production A/Bs on scorer#531 / #532 (Apple M4 10-core, full ≈22 GiB corpus, `--gpu off`):

| N | one-pass wall | wall/N |
|---|---:|---:|
| 1 | 7.72 s | 7.72 s |
| 5 | 23.49 s | 4.70 s |
| 10 | 47.02 s | 4.70 s |
| 25 | 120.89 s | 4.84 s |
| 50 | 244.62 s | 4.89 s |

Post-#531 (scratch reuse) the N=50 median is **206.41 s**. Wall is ~linear in N from N≈5 up: **activation compute dominates; the fixed per-sweep (I/O) share at N≈50 is only ~1–3 %.**

## Provenance

Raised by the performance-candidate survey in stSoftwareAU/NEAT-AI-scorer#536. One candidate per issue; each ships independently on its own evidence — there is no roll-up gate. Small wins accumulate towards the ~10 % cluster-throughput goal.

Burnt ground — do not re-raise: scorer#295 line (#297 activation parallelism, #298 PGO, #299 GPU on production), GPU line (#318/#321/#320/#333/#467), #532 (shared input transpose across the population, −1.37 %), #533 (aggregate-squash flag cache + SIMD feature-detect hoist).

VibeCoderST:
## Summary

Closes #537.

---

🤖 Processed by: VibeCoderST
🏷️ Run id: `vibe-msh2xfx9-32ac82`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -750,7 +750,7 @@ dependencies = [
 
 [[package]]
 name = "neat-core"
-version = "0.8.9"
+version = "0.8.12"
 dependencies = [
  "serde",
  "serde_json",
@@ -1053,9 +1053,9 @@ dependencies = [
 
 [[package]]
 name = "regex-automata"
-version = "0.4.16"
+version = "0.4.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8fcfdb36bda0c880c5931cdc7a2bcdc8ba4556847b9d912bca70bc94708711ad"
+checksum = "ad8553b9b26413251cbf30e620595c7a41b3887f03da04579c0e6b0d6a06b4b2"
 dependencies = [
  "aho-corasick",
  "memchr",
@@ -1076,7 +1076,7 @@ checksum = "19b30a45b0cd0bcca8037f3d0dc3421eaf95327a17cad11964fb8179b4fc4832"
 
 [[package]]
 name = "rust_scorer"
-version = "1.1.53"
+version = "1.1.54"
 dependencies = [
  "bytemuck",
  "clap",
--- a/rust_scorer/Cargo.toml
+++ b/rust_scorer/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rust_scorer"
-version = "1.1.54"
+version = "1.1.55"
 edition = "2024"
 license = "Apache-2.0"
 description = "Native CLI binary for scoring NEAT-AI creatures against training data."
--- a/rust_scorer/src/multi_score.rs
+++ b/rust_scorer/src/multi_score.rs
@@ -29,8 +29,11 @@
 //! straggling task exposes ~1/`k` of the previous tail. `k = 1` is the shipped
 //! default and reproduces the partition above exactly; the total is still
 //! clamped to the host `max_worker_count` RAM ceiling (one `CompiledNetwork`
-//! clone per worker). Raising `k` re-associates a creature's f64 partial sums,
-//! so scores move by rounding-level amounts only.
+//! clone per worker). Raising `k` changes each worker's record count, which
+//! re-associates a creature's f64 partial sums *and* selects a different
+//! 8-record / 4-record / scalar SIMD path in the upstream loss kernels. The
+//! latter dominates: it regroups the f32 activations, so scores move by
+//! f32-rounding amounts (~1e-9 relative in practice), never more.
 
 use std::collections::BTreeMap;
 use std::fs;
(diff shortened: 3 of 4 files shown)


VibeCoderST:
✅ **Vibe Coder released this claim** — host `GRQ-23`, finished 07:09 UTC.

**Progress**
- 06:55 Claimed PR #540 (CI check `Quality Checks` failing)
- 06:55 Diagnosing `Quality Checks` (unknown) — fix attempt 1 of 3

Issue #551, opened 2026-08-10
rust_scorer: capture the x86 Linux tier host-report + neutral baseline (#544/#545)


VibeCoderST:
Follow-up from #545 (sub-issue of #544).

## Problem

#545 shipped the knob report (`rust_scorer --host-report`) and the sweep
harness (`scripts/bench-knob-sweep.sh`) and recorded the **Apple mid** tier
baseline (GRQ-23, M4 / 10 cores / 24 GB) in
`docs/performance-baseline.md` → "#544 fleet knob baseline — 10 August 2026
(Issue #545)".

The **x86 Linux** tier row of that section is still empty: no x86 Linux fleet
host (GRQ-7 / GRQ-12 / GRQ-15 / GRQ-16) is reachable from the unattended worker,
so its detected `HostResources` and neutral baseline timings could not be
captured. Those boxes are the tier most likely to expose a bad default — 4–12
cores, 7.6–15.5 GB RAM, no GPU adapter — so the #544 retune sub-issues are
tuning half blind without it.

The report *code path* is already exercised on the GPU-less Linux CI runner by
`rust_scorer/tests/host_report.rs` (exit 0, valid JSON, complete knob set on
every PR). What is missing is the fleet measurement, which needs a human with
access to one of those machines.

## Scope

On one x86 Linux fleet host, with a local creature pool and corpus:

```bash
cargo build -p rust_scorer --release
./target/release/rust_scorer --host-report

BENCH_SWEEP_CREATURE=/path/to/creatures_dir BENCH_SWEEP_DATA=/path/to/corpus \
  BENCH_SWEEP_VALUES=default ./scripts/bench-knob-sweep.sh
BENCH_SWEEP_CREATURE=/path/to/creatures_dir BENCH_SWEEP_DATA=/path/to/corpus \
  BENCH_SWEEP_GPU=off BENCH_SWEEP_KNOB=NEAT_SCORER_READ_BYTES \
  BENCH_SWEEP_VALUES=default,2097152,8388608,33554432 ./scripts/bench-knob-sweep.sh
```

Append a "Tier: x86 Linux" block to the #545 baseline section in the same shape
as the Apple mid block, replacing the "outstanding" note.

## Acceptance criteria

- [ ] `--host-report` output from one x86 Linux fleet host pasted into
      `docs/performance-baseline.md`.
- [ ] Neutral (`BENCH_SWEEP_VALUES=default`) baseline timings for that host.
[...]

Issue #545, opened 2026-08-10
rust_scorer: host knob-report and fleet benchmark harness for self-tuning (#544)


stservice:
Sub-issue of #544.

## Problem

#544 requires benchmark-first retuning of every host-tuned knob, but there is
today no way to (a) see what the scorer actually detected and chose on a given
fleet host, or (b) sweep a knob across its candidate values on the production
scoring path. `docs/performance-baseline.md` sweeps were hand-driven per issue
(#307, #467, #529) and are not repeatable across the 20+ machine fleet.

Without this, every other #544 sub-issue has to re-invent its own measurement
rig, and "before/after numbers in the PR" cannot be produced consistently.

## Scope

1. **Knob report.** Add a diagnostic that prints the detected `HostResources`
   and every resolved knob as one JSON object — logical CPUs, physical RAM,
   plus the resolved `default_worker_count`, `max_worker_count`,
   `max_read_bytes`, `default_training_read_bytes_for(record_bytes)` and the
   GPU scratch budget, each tagged with whether it came from the built-in
   default or an env override. Surface it on the existing CLI (a flag such as
   `--host-report`, following the established flag style in
   `rust_scorer/src/cli.rs`).
2. **Knob sweep harness.** A script under `scripts/` that runs the production
   scoring path at a caller-supplied list of values for one knob and emits
   median wall-clock per value, in the same shape the existing
   `scripts/bench-shallow-gpu.sh` A/B produces. It must skip cleanly when the
   local corpus/creature inputs are unset, exactly as the production benches do
   (this public repo ships no production creature — Issue #448).
3. **Fleet baseline.** Capture the report + a single-knob-neutral baseline run
   on at least one host from each real tier and record it in
   `docs/performance-baseline.md`.

## Real fleet tiers (from GRQ-health `docs/index.json`)

| Tier | Example hosts | CPU | Logical cores | Reported RAM | GPU cores |
|---|---|---|---|---|---|
| Apple large | Mac-Ultra-M2 | M2 Ultra | 24 | 64 GB | 60 |
| Apple mid-high | GRQ-25 | M4 Pro | 12 | 24 GB | 16 |
| Apple mid | GRQ-23, GRQ-21, GRQ-18, GRQ-19, GRQ-22 | M4 | 10 | 16–24 GB | 10 |
| Apple older mid | GRQ-11, GRQ-3 | M1 Max / M1 Pro | 10 | 32 / 16 GB | 24 / 16 |
[...]

Issue #544, opened 2026-08-09
rust_scorer: self-tune read, worker and GPU knobs to detected host hardware


nleck:
Tuning for an M4 or better Mac . Looking for ways to eek out some more performance

## Current Understanding

The user wants the Rust scorer to be fully self-tuning: at startup it should detect the host's memory, CPU and GPU and choose every performance knob itself — no per-host env-var recipes. The fleet is a mixed family of machines (GRQ-health `docs/index.json`), with M-series Macs doing the vast majority of evolution work and future machines expected to be modern Macs or similar. Benchmark-first rules apply to any retuning.

### Accepted scope so far

- All host-tuned knobs are in scope: read-chunk defaults (`rust_scorer/src/read_tuning.rs`), worker counts and ceilings, and the GPU scratch budget (`rust_scorer/src/host_resources.rs`, consumed by `rust_scorer/src/gpu/`).
- Deliverable is retuned built-in self-tuning defaults; env overrides stay as emergency escape hatches only, never as per-host configuration.
- Target is the whole fleet family, not one host: M1 (8 GiB) through M4 Pro (24 GiB) Macs, one M2 Ultra (64 GiB), plus older x86 Linux boxes that must not regress.
- Benchmark-first workflow per knob: baseline before any change, before/after numbers in any PR, `negative-result` close for any knob showing no meaningful gain.

### Open questions

- None.

### Assumptions

- "Check the memory/cpu/gpu" means extending the `HostResources` probe (today it senses only logical CPUs and physical RAM) with GPU capability sensing; tuning remains deterministic detection-based defaults.
- Whether a tier-table heuristic or a startup micro-calibration serves a given knob better is a design decision for planning, settled by benchmarks.
- Tier boundaries should reflect the real fleet: no current M4 host has 64 GiB+ RAM (they run 16–24 GiB), so today's 64 GiB+ branches only ever fire on the M2 Ultra.
- Benchmarks run on the production scoring path on fleet machines (primarily M-series Macs).

### Related open issues

- None. (#544 is the only open issue; the recent perf chain #529, #534, #537, #539 is all closed.)

VibeCoderST:
✅ **Vibe Coder released this claim** — host `GRQ-23`, finished 00:56 UTC.

VibeCoderST:
## Grill-Me Round 1

[...]

PR #559, opened 2026-08-10
rust_scorer: drop the dead 256 MiB read ceiling and bound the reader footprint (Issue #549)

### LINKED ISSUE #549: rust_scorer: retune read-chunk defaults to real fleet tiers and reader count (#544)
Sub-issue of #544.

## Problem

The read-chunk defaults in `rust_scorer/src/read_tuning.rs` were tuned in #307
and given RAM awareness in #534, but the tier table does not match the real
fleet and contains an unreachable branch.

**1. The 256 MiB ceiling is dead.** `host_resources::max_read_bytes` allows
256 MiB on hosts with ≥ 64 GiB RAM, but `default_training_read_bytes_for`
selects `MAX_READ_BYTES` (64 MiB) for that same tier:

```rust
Some(ram) if ram >= 64 * GIB => MAX_READ_BYTES,   // 64 MiB, read_tuning.rs:40
```

So the 256 MiB ceiling is only reachable by setting the env var by hand — which
#544 explicitly rules out as a configuration mechanism. And the ≥ 64 GiB tier
fires on exactly **one** fleet host (Mac-Ultra-M2), so it has never been
measured against the machines doing the work.

**2. The tiers do not discriminate where the fleet actually lives.** Every
Apple host from an 8 GB M1 to a 24 GB M4 Pro that reads production-width
records (~9848 bytes/record, above `LARGE_RECORD_BYTES_THRESHOLD`) lands on the
same `LARGE_RECORD_DEFAULT_READ_BYTES` = 32 MiB, subject only to the RAM cap.
The M4/M4 Pro/M1-Max class spans a 3× range in memory bandwidth and GPU
capability with no differentiation.

**3. The chunk size interacts with the #529 reader count.** With one reader per
CPU and 26 concurrent `.bin` shards in production, peak resident read buffer is
`readers × chunk`. On a 10-core, 16 GB M4 that is 10 × 32 MiB = 320 MiB of read
buffer — a budget the current per-knob tiering never accounts for, because
`read_tuning.rs` sizes the chunk without knowing the reader count.

## Scope

- Re-derive the read-chunk default table against the real fleet tiers, sweeping
  chunk size on the production scoring path per tier.
- Resolve the dead 256 MiB ceiling: either raise the ≥ 64 GiB default to use it
  or drop the ceiling to what the defaults can actually select — whichever the
  M2 Ultra numbers support.
- Make the default aware of the concurrent reader count so the aggregate
  `readers × chunk` footprint stays within a sane fraction of host RAM, rather
  than sizing each reader in isolation.
- Keep the small-record path (`< 8000` bytes/record) and the existing
  `NEAT_SCORER_READ_BYTES` override and clamp semantics intact.

## Benchmark plan

Sweep chunk size per tier on the `fused_multi_file` group at
`BENCH_FUSED_FILES=26` and on the production forward-only path, at the shipped
reader count — the #307 sweep only measured a single sequential reader, which
is no longer the default since #529. Cover M2 Ultra, M4 Pro, M4, M1-class and
one x86 Linux control.

## Acceptance criteria

- [ ] No unreachable tier remains: every ceiling in `max_read_bytes` is
      selectable by a built-in default on some real host, or is removed.
- [ ] Aggregate read-buffer footprint at the shipped reader count is bounded
      and unit-tested over synthetic hosts.
- [ ] Before/after medians per tier in the PR and appended to
      `docs/performance-baseline.md`, at a stated `BENCH_SCORING_BYTES`.
- [ ] x86 Linux control shows no regression.
- [ ] Env override and record-alignment behaviour unchanged (the result is
      still a whole number of records, at least one record wide).
- [ ] Any tier showing no meaningful gain keeps its current default and the
      negative result is recorded.

## Failure Detection

- **Unit tests in `rust_scorer/src/read_tuning.rs` (`#[cfg(test)]` module), run by `cargo test` in the scorer CI workflow — earliest detection point.** Add/extend cases so a regression fails at PR time:
  - `no_unreachable_ceiling`: for every RAM tier in `host_resources::max_read_bytes`, assert some synthetic host makes `default_training_read_bytes_for` select a value that reaches that ceiling (or the ceiling no longer exists). Reintroducing a dead branch like the current `≥ 64 GiB → MAX_READ_BYTES` fails this test.
  - `aggregate_footprint_bounded`: over a grid of synthetic hosts (8 GB M1 through M2 Ultra, plus a 10-core/16 GB M4 case), assert `readers × chunk` stays within the chosen fraction of host RAM. The specific regression from the issue (10 × 32 MiB = 320 MiB on a 16 GB M4) must be a failing input before the fix and a passing one after.
  - Existing override/clamp tests: `NEAT_SCORER_READ_BYTES` env override, whole-record alignment, and at-least-one-record minimum must pass unchanged; any behavioural drift in those knobs fails the suite.
- **Benchmark gate at PR time:** the `fused_multi_file` group at `BENCH_FUSED_FILES=26` and the forward-only path, run at the shipped reader count per tier. A retuned default that is slower than the current default on any measured tier (M2 Ultra, M4 Pro, M4, M1-class, x86 Linux control) is a blocking regression; medians go into the PR and `docs/performance-baseline.md`, so later drift is detectable by re-running the sweep against the recorded baseline.
- **Fleet-side detection if a bad default ships:** scoring-cycle wall-time in the GRQ node logs rises on the affected tier, and on memory-tight hosts an oversized aggregate read buffer surfaces as heap-pressure DEFERs in the evolution heap ledger — both visible in the nightly GRQ-logs sweep. This is the last-resort signal; the unit tests and per-tier benchmark comparison above should catch any regression before merge.

VibeCoderST:
# Retune read-chunk defaults against the real fleet and the reader count (Issue #549)

## Summary

Removes the dead 256 MiB read ceiling and makes the read-chunk default aware of
the concurrent reader count, so the `readers × chunk` footprint is bounded by
host RAM rather than by a clamp the defaults could never reach. The chunk-size
*retune* itself is a recorded negative result: no fleet tier's shipped chunk
size changes. Closes #549.

Three changes:

1. **`host_resources::max_read_bytes` returns one 64 MiB ceiling on every
   host.** The old `≥ 64 GiB → 256 MiB` branch was unreachable —
   `read_tuning::default_training_read_bytes_for` selected `MAX_READ_BYTES`
   (64 MiB) for that same tier — so the only way to reach 256 MiB was to export
   `NEAT_SCORER_READ_BYTES` by hand, which #544 rules out as a configuration
   mechanism. The surviving ceiling *is* selectable: a ≥ 64 GiB host reading
   production-width records defaults to exactly 64 MiB.
2. **The aggregate footprint is now a policy in `read_tuning`, not a side
   effect of a clamp.** `aggregate_read_budget_bytes(host)` =
   `min(64 MiB, RAM / 64)`; `per_reader_read_bytes` gives each reader its share
   of it. `stream_score::per_reader_read_buf_len` delegates there, so the
   default *and* an env override both pass through one bound.
3. **No tier's chunk grew.** The RAM share is applied as a `min` — a guard, not
   a raise — because giving each reader the full 32 MiB default measured ~50 %
   slower at the shipped reader count (below). The one behavioural move is on
   ≥ 64 GiB Macs, whose aggregate drops from the old 256 MiB clamp to 64 MiB:
   the direction that measured faster.

```mermaid
flowchart LR
    A[chunk: env override<br/>or adaptive default] --> B["aggregate budget<br/>min(64 MiB, RAM / 64)"]
    B --> C[share = budget / readers]
    C --> D["chunk = min(chunk, share)"]
    D --> E[round down to whole records<br/>at least one record]
```

## Evidence

Backend/CLI change — no web interface to screenshot. Evidence is Criterion
medians plus the unit tests below.

**Host:** Apple M4, 10 logical cores, 24 GB, macOS 25.6. **Not quiescent** —
the unattended worker holds ~1 core (load average 5–9), so only *interleaved*
comparisons are trusted. Corpus `BENCH_SCORING_BYTES=200000000`,
`BENCH_FUSED_FILES=26`, production record width (2461 in / 1 out → 9848 B),
bench `fused_multi_file/file_workers/auto` (10 readers over 26 shards).

### A/B — per-reader chunk, interleaved (5 pairs, same binary)

| Pair | Shipped 6 706 488 B (6.40 MiB) | Isolated 32 MiB/reader | Δ |
|---:|---:|---:|---:|
| 1 | 27.670 ms | 40.700 ms | +47 % |
| 2 | 26.191 ms | 39.358 ms | +50 % |
| 3 | 27.548 ms | 42.019 ms | +53 % |
| 4 | 27.863 ms | 42.049 ms | +51 % |
| 5 | 27.151 ms | 47.859 ms | +76 % |
| **median** | **27.548 ms** | **42.019 ms** | **+53 %** |

Sizing each of ten concurrent readers at the full 32 MiB default is ~50 %
slower, 5 of 5 pairs — so the aggregate budget is not raised to a RAM share
that would allow it.

### Sweep — per-reader chunk at 10 readers (3 rounds)

| Per-reader chunk | Round 1 | Round 2 | Round 3 | Median |
|---|---:|---:|---:|---:|
| 2 MiB | 24.298 ms | 24.698 ms | 26.497 ms | **24.698 ms** |
| 4 MiB | 25.359 ms | 25.876 ms | 24.907 ms | 25.359 ms |
| 6.40 MiB (shipped) | 27.190 ms | 29.050 ms | 22.284 ms | 27.190 ms |
| 12 MiB | 30.367 ms | 31.177 ms | 24.679 ms | 30.367 ms |
| 16 MiB | 32.299 ms | 36.631 ms | 25.375 ms | 32.299 ms |

Rounds 1–2 rank smaller-is-faster monotonically; round 3 (a quieter interval)
inverts the order and its fastest single sample is the shipped size. A ~9 % lead
one round in three contradicts, on a contended host, is not evidence for a
fleet-wide default change — recorded as an open lead in
`docs/performance-baseline.md`.

### Shipped defaults, before vs after

Unchanged on every fleet host below 64 GiB — asserted for the whole fleet grid
by `no_fleet_tier_gets_a_bigger_chunk_than_before`:

| Tier | Readers | Chunk before | Chunk after |
|---|---:|---:|---:|
| M1 8 GB | 8 | 7.99 MiB | 7.99 MiB |
| x86 Linux 16 GB | 8 | 7.99 MiB | 7.99 MiB |
| M4 / M1 Pro 16 GB | 10 | 6.40 MiB | 6.40 MiB |
| M4 24 GB (measured) | 10 | 6.40 MiB | 6.40 MiB |
| M4 Pro 24 GB | 12 | 5.33 MiB | 5.33 MiB |
| M2 Ultra 64/192 GB | 24 | 10.66 MiB | 2.66 MiB |

Because the sizes are byte-identical on every measured tier, the
`fused_multi_file` group before/after (`--baseline before-549`) shows only host
noise: `auto` −4.0 %, `file_workers/1` −0.8 %, `/2` −16.5 %, `/4` +11.5 %,
`/8` +8.0 % — inconsistent directions on an identical code path.

### Acceptance criteria not met

* **x86 Linux control and the M2 Ultra / M4 Pro / M1-class sweeps** were not
  captured: no such host is reachable from this worker (the same gap Issue #551
  tracks for the #545 fleet baseline). The mitigation is that the shipped chunk
  size is provably unchanged on every tier below 64 GiB, so there is nothing to
  regress; the ≥ 64 GiB move is a tightening in the measured-faster direction.

## Test Plan

New/changed unit tests in `rust_scorer/src/read_tuning.rs` (`cargo test`, run by
the scorer CI workflow):

- `no_unreachable_ceiling` — every ceiling `host_resources::max_read_bytes` can
  return over the fleet grid must be selected by a built-in default on some real
  host. **Failed before the fix**: "no fleet host's built-in default selects the
  268435456 B read ceiling".
- `aggregate_footprint_bounded` — over the fleet grid (8 GB M1 … 192 GB M2
  Ultra, including the 10-core / 16 GB M4), `readers × chunk` stays inside the
  budget, the budget stays inside `RAM / 64`, and the chunk stays record-aligned
  and ≥ one record. **Failed before the fix**: "M1 Pro 16 GB: 10 readers x
  33552136 B = 335521360 B exceeds the 268435456 B budget" — the exact 320 MiB
  regression from the issue.
- `reader_unaware_sizing_blows_the_budget_on_a_16_gib_m4` — pins that the
  reader-unaware default is the failing input and the reader-aware one is not.
- `no_fleet_tier_gets_a_bigger_chunk_than_before` — encodes the negative result:
  no tier's chunk grows, and every tier below 64 GiB keeps exactly the size it
  shipped with.
- `per_reader_share_never_grows_a_request`, `unknown_ram_keeps_the_pre_549_shared_budget`.
- `host_resources::every_host_shares_one_read_ceiling`.

Modified (business-logic change, not weakened):

- `host_resources::large_mac_raises_read_and_scratch_ceilings` and
  `exact_24_gib_and_64_gib_unchanged` — the 256 MiB read assertions become
  64 MiB (the scratch assertions are untouched).
- `stream_score::per_reader_read_buf_shares_one_total_budget` — asserts against
  the new aggregate budget instead of the fixed ceiling, and gains a "never
  grows the request" assertion.

Unchanged and still passing: the `NEAT_SCORER_READ_BYTES` override, clamp and
whole-record-alignment tests/doctests, and the full suite (256 lib tests +
integration + doctests) via `./quality.sh`.

## Milestone
This PR is part of the **#544 rust_scorer: self-tune read, worker and GPU knobs to detected host hardware** milestone and targets the `milestone/544-rust-scorer-self-tune-read-worker-and-gpu-knob` feature branch.

---

🤖 Processed by: VibeCoderST
🏷️ Run id: `vibe-msmwq6q7-f5c08d`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,21 @@ section to the released version with its date.
 
 ### Fixed
 
+- **Read-chunk ceiling and aggregate reader footprint (Issue #549).**
+  `host_resources::max_read_bytes` now returns **one** 64 MiB ceiling on every
+  host: the ≥ 64 GiB / 256 MiB tier was unreachable by any built-in default —
+  `read_tuning` capped even a 192 GiB Mac at 64 MiB — so it could only be
+  reached by exporting `NEAT_SCORER_READ_BYTES` by hand, which Issue #544 rules
+  out as a configuration mechanism. The `readers × chunk` footprint is now
+  bounded by `read_tuning::aggregate_read_budget_bytes` = `min(64 MiB,
+  RAM / 64)`, which both the default and an env override pass through, instead
+  of each reader being sized in isolation against a clamp. The RAM share only
+  ever *tightens*, so **no fleet tier's chunk size changed**: giving each of ten
+  readers the full 32 MiB default measured ~50 % slower on M4 / 24 GB at the
+  shipped reader count (5 of 5 interleaved pairs) — the retune half is a
+  recorded `negative-result` in `docs/performance-baseline.md`. Env override,
+  whole-record alignment and the one-record minimum are unchanged.
+
 - **PR auto-format syncs `Cargo.lock` to latest neat-core (Issue #542).** The
   auto-format job now runs `cargo update -p neat-core` after `cargo fmt`, so
   every PR refreshes the path-dependency lock entry against the checked-out
--- a/docs/performance-baseline.md
+++ b/docs/performance-baseline.md
@@ -172,6 +172,95 @@ timings). Tracked in
 two commands above on one x86 Linux host and append a "Tier: x86 Linux" block
 in the same shape.
 
+## Read-chunk retune at the shipped reader count — 10 August 2026 (Issue #549)
+
+**Dead ceiling removed and the aggregate footprint bounded; the chunk-size
+*retune* is a recorded `negative-result`.** The 256 MiB read ceiling for
+≥ 64 GiB hosts was unreachable by any built-in default and is gone, and the
+`readers × chunk` budget is now a bounded share of physical RAM in
+[`read_tuning`](../rust_scorer/src/read_tuning.rs) instead of a fixed clamp read
+out of `host_resources`. **No fleet host's shipped chunk size changes** — the
+A/B below is why.
+
+### Host
+
+| Field | Value |
+|---|---|
+| Machine | Apple M4, 10 logical cores (4P + 6E), 24 GB, macOS 25.6 |
+| Corpus | `BENCH_SCORING_BYTES=200000000`, `BENCH_FUSED_FILES=26` |
+| Record width | `BENCH_SCORING_INPUTS=2461`, `BENCH_SCORING_OUTPUTS=1` → 9848 B/record (production width) |
+| Bench | `fused_multi_file/file_workers/auto` — the shipped reader count (10 readers over 26 shards) |
+| Contention | Not quiescent: the unattended worker holds ~1 core (load average 5–9), so only interleaved comparisons are trusted |
+
+### A/B — per-reader chunk, interleaved (5 pairs, alternating)
+
+The two arms are the *same* binary at two per-reader chunk sizes: the shipped
+6 706 488 B (64 MiB aggregate ÷ 10 readers, record-aligned) against the
+32 MiB the isolated large-record default would give each reader.
+
+| Pair | Shipped 6 706 488 B (6.40 MiB) | Isolated 32 MiB/reader | Δ |
+|---:|---:|---:|---:|
+| 1 | 27.670 ms | 40.700 ms | +47 % |
+| 2 | 26.191 ms | 39.358 ms | +50 % |
+| 3 | 27.548 ms | 42.019 ms | +53 % |
+| 4 | 27.863 ms | 42.049 ms | +51 % |
+| 5 | 27.151 ms | 47.859 ms | +76 % |
+| **median** | **27.548 ms** | **42.019 ms** | **+53 %** |
+
+Sizing each of ten concurrent readers at the full 32 MiB default is **~50 %
+slower**, 5 of 5 interleaved pairs. Raising the aggregate budget to a share of
+RAM (which is what would let every reader take 32 MiB) therefore does not ship:
+the RAM share is applied as a `min` against the existing ceiling, exactly as the
+GPU scratch share is in Issue #548.
+
+### Sweep — per-reader chunk at 10 readers (3 rounds, same order)
+
+| Per-reader chunk | Round 1 | Round 2 | Round 3 | Median |
+|---|---:|---:|---:|---:|
+| 2 MiB | 24.298 ms | 24.698 ms | 26.497 ms | **24.698 ms** |
+| 4 MiB | 25.359 ms | 25.876 ms | 24.907 ms | 25.359 ms |
+| 6.40 MiB (shipped) | 27.190 ms | 29.050 ms | 22.284 ms | 27.190 ms |
+| 12 MiB | 30.367 ms | 31.177 ms | 24.679 ms | 30.367 ms |
+| 16 MiB | 32.299 ms | 36.631 ms | 25.375 ms | 32.299 ms |
+
+Rounds 1 and 2 rank smaller-is-faster monotonically; round 3 (a quieter
+interval — every size ran faster) inverts the order, and its fastest single
+sample is the *shipped* size. A ~9 % lead for 2 MiB that one round out of three
+contradicts, on a host under known contention, is not evidence for a fleet-wide
+default change — and the other tiers named in Issue #549 (M2 Ultra, M4 Pro,
+M1-class, and the x86 Linux control) are not reachable from this worker, so no
+per-tier retune ships. **Open lead:** a quiescent multi-tier sweep of chunk
+sizes *below* the shipped one, in the same shape as the table above.
+
+### What shipped
+
+* `host_resources::max_read_bytes` returns **one** 64 MiB ceiling on every host.
+  The ≥ 64 GiB / 256 MiB tier could only be reached by exporting
+  `NEAT_SCORER_READ_BYTES` by hand, which
+  [#544](https://github.com/stSoftwareAU/NEAT-AI-scorer/issues/544) rules out as
+  a configuration mechanism.
+* `read_tuning::aggregate_read_budget_bytes` owns the `readers × chunk` bound:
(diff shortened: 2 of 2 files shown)


nleck:
Please fix merge issues

nleck (approved):

VibeCoderST:
Claiming PR feedback comment 5239844465 for worker `GRQ-23-19412`.

VibeCoderST:
✅ **Vibe Coder released this claim** — host `GRQ-23`, finished 12:23 UTC.

VibeCoderST:
I fixed the issues from this feedback locally but failed to push the changes. Please check the branch status.

PR #573, opened 2026-08-18
Refuse rust_scorer version downgrades vs Develop (Issue #567)

### LINKED ISSUE #567: Refuse rust_scorer version downgrades vs Develop
## Problem

`scripts/version-increment.sh --already-bumped` / `--run` treat any `current != base` as success, including **downgrades**. Scorer version is what GRQ pins / rebuilds for `rust_scorer`.

## Acceptance criteria

1. A PR whose package version is **strictly less than** `origin/Develop` fails CI (not skipped / not treated as “already bumped”).
2. Equal versions may still auto-patch-bump as today.
3. Ahead versions are accepted without forcing another bump.
4. A unit / script WHAT-test covers at least: behind → fail, equal/ahead → pass.
5. Document the rule briefly in CONTRIBUTING (or the existing version-increment notes).

## Reference implementations

- TypeScript / `deno.json`: [NEAT-AI `test/ci/PackageVersionNoDowngrade.ts`](https://github.com/stSoftwareAU/NEAT-AI/blob/Develop/test/ci/PackageVersionNoDowngrade.ts) and the `sort -V` guard in [`update-package-version.yml`](https://github.com/stSoftwareAU/NEAT-AI/blob/Develop/.github/workflows/update-package-version.yml)
- Rust / `Cargo.toml`: [NEAT-AI-core `scripts/check-version-bump.sh`](https://github.com/stSoftwareAU/NEAT-AI-core/blob/Develop/scripts/check-version-bump.sh) (rejects `head < base`)

## Why

GRQ and sibling rebuilds key off crate / package versions. A merge conflict that silently takes Develop’s older version looks “already different” to today’s `CURRENT != BASE` skip logic and ships a **downgrade** — the same failure mode that forced NEAT-AI’s floor restore to 6.5.0 / 6.6.0.

## Suggested touchpoints

- `scripts/version-increment.sh` (`--already-bumped` and `run`)
- `.github/workflows/version-increment.yml`
- Exit non-zero (or bump from base) when `current < base`

stservice:
# Refuse `rust_scorer` version downgrades vs the base ref (Issue #567)

## Summary

`scripts/version-increment.sh` treated **any** `current != base` as "already
bumped", so a branch whose `rust_scorer` version was *below* `origin/Develop` —
the shape a badly resolved merge conflict produces — was silently skipped and
shipped a **downgrade** to the downstream consumers that pin and rebuild
`rust_scorer` by version.

The script now compares the two versions instead of merely testing them for
inequality:

- **Behind → fail loud.** Both `--already-bumped` and `--run` exit **3** with
  `Error: version downgrade refused — branch version X is behind base Y …`.
  `--run` writes nothing, so the manifest is left as the author committed it.
- **Equal → auto-patch-bump**, exactly as before.
- **Ahead → accepted**, with no second bump forced.
- Comparison is **numeric per component** (`0.10.0` is ahead of `0.9.9`, which a
  string compare gets wrong), and a pre-release suffix sorts below the bare
  release of the same triple (`0.5.4-rc1` < `0.5.4`), matching semver.
- A version that is not semver `X.Y.Z` exits **2** with a clear message rather
  than being mistaken for a bump.

No workflow YAML change is required, and none is included (the worker holds no
`workflow` OAuth scope — see the CONTRIBUTING [Human
escalation](../../../CONTRIBUTING.md#human-escalation) contract). The existing
wiring already fails CI on a downgrade: the guard job's `--already-bumped` exits
non-zero, so `should_bump=true` and the `bump` job runs `--run`, which exits 3
and fails the job before anything is committed or pushed. That guard → bump
sequence is pinned by a test.

Closes #567.

## Evidence

Backend/CLI change — no web interface to screenshot. Evidence is the BATS suite
(`bats tests/scripts/version_increment.bats`), 20/20 passing, with the six new
downgrade cases failing before the change and passing after:

```text
ok  9 already-bumped? refuses a patch downgrade against the base ref
ok 10 already-bumped? refuses a minor downgrade against the base ref
ok 11 already-bumped? accepts an ahead version (no re-bump forced)
ok 12 already-bumped? compares components numerically, not as strings
ok 13 already-bumped? treats a pre-release of the base version as behind it
ok 14 run refuses a downgrade and leaves the manifest untouched
ok 15 run accepts an ahead version without forcing another bump
ok 16 run still bumps when the branch version equals the base
ok 17 a non-semver branch version fails loud rather than being called a bump
ok 18 the workflow guard/bump sequence fails CI on a downgrade
```

Decision flow now applied by both `--already-bumped` and `--run`:

```mermaid
flowchart TD
    A[Read branch version + base ref version] --> B{Base ref reachable?}
    B -- no --> C[Conservative: not bumped]
    B -- yes --> D{Compare branch vs base}
    D -- "branch &lt; base" --> E[Exit 3<br/>downgrade refused — CI red]
    D -- "branch == base" --> F[Patch-bump once]
    D -- "branch &gt; base" --> G[Skip — already bumped]
```

## Test Plan

Added to `tests/scripts/version_increment.bats` (all exercise the real script
against temporary git repositories, asserting exit codes, messages and manifest
side-effects):

- `already-bumped? refuses a patch downgrade against the base ref` — 0.5.3 vs
  0.5.4 → exit 3, message names both versions.
- `already-bumped? refuses a minor downgrade against the base ref` — 0.4.9 vs
  0.5.4 → exit 3.
- `already-bumped? accepts an ahead version (no re-bump forced)` — 0.6.0 → exit 0.
- `already-bumped? compares components numerically, not as strings` — 0.10.0 vs
  0.5.4 → exit 0 (a lexical compare would call this a downgrade).
- `already-bumped? treats a pre-release of the base version as behind it` —
  0.5.4-rc1 vs 0.5.4 → exit 3.
- `run refuses a downgrade and leaves the manifest untouched` — exit 3 and the
  manifest still reads 0.5.3.
- `run accepts an ahead version without forcing another bump` — skip, 0.10.0
  preserved.
- `run still bumps when the branch version equals the base` — 0.5.4 → 0.5.5.
- `a non-semver branch version fails loud rather than being called a bump` —
  exit 2, "semver" in the message.
- `the workflow guard/bump sequence fails CI on a downgrade` — replays the
  workflow's guard (`--already-bumped`) → bump (`--run`) ordering and asserts
  both steps end non-zero.

Existing tests are unchanged and still pass; the only behavioural change to a
previously documented path is that a *behind* version, which used to be reported
as "already bumped", is now refused.

Docs updated: the no-downgrade rule is documented in the CONTRIBUTING
[Pull request workflow](../../../CONTRIBUTING.md#pull-request-workflow) section,
the script's own `--help` now lists the exit codes, and `CHANGELOG.md` carries
an `[Unreleased]` entry.

---

🤖 Processed by: stservice
🏷️ Run id: `vibe-msyib942-619600`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,19 @@ section to the released version with its date.
 
 ### Added
 
+- **`rust_scorer` version downgrades are refused (Issue #567).**
+  `scripts/version-increment.sh` now compares the branch version with the base
+  ref numerically instead of treating any `current != base` as "already
+  bumped": a version strictly below the base exits **3** with a
+  `version downgrade refused` message from both `--already-bumped` and
+  `--run`, so a merge conflict that restores Develop's older version fails CI
+  instead of shipping a downgrade to a downstream `rust_scorer` pin. Equal
+  versions still auto-patch-bump; ahead versions are accepted without a bump;
+  `0.10.0` is correctly ahead of `0.9.9` and a pre-release suffix sorts below
+  the bare release of the same triple. Covered by
+  `tests/scripts/version_increment.bats` (behind → fail, equal/ahead → pass)
+  and documented in the CONTRIBUTING PR-workflow section.
+
 - **Single `input`/`output` ≥ 1 guard on every scoring path, with regression
   tests (Issue #571).** `rust_scorer/src/creature_width.rs` is now the one
   place the observation-width rule is worded
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -170,6 +170,20 @@ automatically bumps the patch component of `rust_scorer`'s version in
 branch. Because the version is bumped automatically, the `CHANGELOG.md` is
 the human-readable record of *what* changed — please keep it current.
 
+**Version downgrades are refused (Issue #567).** Downstream consumers pin and
+rebuild `rust_scorer` by version, so a branch version *below* the base ref must
+never ship. [`scripts/version-increment.sh`](./scripts/version-increment.sh)
+compares the branch version with the base numerically (`0.10.0` is ahead of
+`0.9.9`; a pre-release suffix sorts below the bare release of the same
+triple) and both `--already-bumped` and `--run` exit **3** with a
+`version downgrade refused` message when the branch is behind — a downgrade
+is never mistaken for "already bumped", and CI goes red rather than pushing
+it. Equal versions still auto-patch-bump, and an ahead version is accepted
+without forcing a second bump. If a merge conflict restores Develop's older
+version, set `[package].version` in
+[`rust_scorer/Cargo.toml`](./rust_scorer/Cargo.toml) to at least the base
+version and push again.
+
 ## Performance Task Workflow
 
 This section is the **single home** for the project's performance-change rules.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -750,7 +750,7 @@ dependencies = [
 
 [[package]]
 name = "neat-core"
-version = "0.9.4"
+version = "0.9.5"
 dependencies = [
  "serde",
  "serde_json",
@@ -1076,7 +1076,7 @@ checksum = "19b30a45b0cd0bcca8037f3d0dc3421eaf95327a17cad11964fb8179b4fc4832"
 
 [[package]]
 name = "rust_scorer"
-version = "1.1.62"
+version = "1.1.63"
 dependencies = [
  "bytemuck",
  "clap",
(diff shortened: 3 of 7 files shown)


stservice:

- `Cargo.lock`

```
  bump: cc -> 1.4.3
  bump: clap -> 4.6.6
  bump: clap_builder -> 4.6.6
  bump: futures-core -> 0.3.34
  bump: futures-task -> 0.3.34
  bump: futures-util -> 0.3.34
  bump: pkg-config -> 0.3.34
  fail: syn -> 2.0.119 (cargo update rejected)
  bump: thiserror -> 2.0.20
  bump: xml-rs -> 0.8.29
  bump: zerocopy -> 0.8.56
external: 16 bumped, 0 deferred, 1 failed

```

</details>

PR #580, opened 2026-08-23
Directory mode rejects forwardOnly=false — the recurrent maths already exists in cli.rs, only the batch loop hard-codes the flag (Issue #579)

### LINKED ISSUE #579: Directory mode rejects forwardOnly=false — the recurrent maths already exists in cli.rs, only the batch loop hard-codes the flag
Raised so the sibling NEAT-AI repo can delete its TypeScript/WASM dataset-scoring path and make `rust_scorer` the single scoring implementation (DRY — two implementations of the same maths can and do disagree).

## The blocker

Single-creature mode **already scores recurrent creatures natively**. `rust_scorer/src/cli.rs:520`:

```rust
let use_fused_stream = creature.forward_only;
```

`forwardOnly=false` takes the `else` branch at `cli.rs:563-604` — a `TrainingDataIterator` per-record loop calling `cost::accumulate_cost_sum(..., forward_only = false)`. Verified by running the binary: valid JSON, `"forwardOnly": false`, `"trainingReadBackend": "record_iterator"`, exit 0. A genuine back edge changes the answer (error `0.0311` with `forwardOnly:true` vs `4.4e-16` with `forwardOnly:false` on the same topology and corpus), so the flag is load-bearing rather than cosmetic.

Directory mode rejects them at load time — `rust_scorer/src/multi_score.rs:234-239`:

```rust
if !creature.forward_only {
    return Err(format!("Creature '{}' has forwardOnly=false; multi-creature directory mode requires forwardOnly=true for every creature", path.display()));
}
```

Runtime output (exit 1, **no JSON at all**):

```
Error: Creature '…/rec.json' has forwardOnly=false; multi-creature directory mode requires forwardOnly=true for every creature
```

Documented at `README.md:136`, pinned by `tests/directory_mode_tdd.rs:527`.

## It is an unimplemented case, not a design limit

The guard exists because the directory hot loop **hard-codes the flag**:

- `multi_score.rs:757-764` — the CPU per-chunk `par_iter` worker passes literal `true` as the last argument to `accumulate_cost_sum`.
- `multi_score.rs:682-689` — same in the empty-chunk cost-dispatch probe.

Without the guard a recurrent creature would be scored with state leaking across records — silently wrong. So it is a correctness stopgap for a hard-coded boolean.

Everything downstream is already flag-driven:

- `rust_scorer/src/cost.rs:225-289` — `accumulate_cost_sum(..., forward_only)` forwards the flag to `neat_core::loss::*_sum_batch_packed` for **all eight** costs.
- `NEAT-AI-core/neat-core/src/loss.rs:345-349` — with `forward_only = false`, `packed_record_scan` calls `network.reset_state()` before every record, so records stay **independent**. That matters: partitioning a chunk's records across Rayon workers remains valid for recurrent creatures, so no restructuring of the worker-split machinery is needed.
- **The GPU shaders already implement the recurrent semantics.** `shaders/forward_mse_batched.wgsl:265-267` and `forward_mse_scratch.wgsl:259-260` zero every non-input activation per `(creature, record)` thread — a GPU thread never carries state between records. No shader change required.

## Proposed work

1. Delete the guard at `multi_score.rs:234-239`.
2. Build a per-worker flag vector alongside the existing `worker_creature_idx` (`multi_score.rs:~635-641`) and pass `worker_forward_only[worker_idx]` instead of the literal `true` at `:763` (and the `:682` probe).
3. Confirm `gpu_directory_probe_for_dir` (`:1018`) and `gpu_directory_compatible` (`:976`) need no change — they only classify neuron counts.
4. Extend `tests/directory_mode_tdd.rs` and `tests/cost_parity.rs`: a directory containing a recurrent creature must produce byte-identical scores to the single-creature recurrent path, including a **mixed** forward-only + recurrent batch.
5. Update `README.md:136` and the CLI doc comment.

**Expected cost:** recurrent creatures lose the SIMD batch kernels — `neat-core/src/loss.rs:422` and `:436` gate the 8-way and 4-way paths on `forward_only` — so they fall to the scalar `packed_record_scan`. Several-× slower **for those creatures only**; forward-only creatures in the same batch are unaffected. `reset_state()` adds O(num_neurons) stores per record (~3000 f32 writes against ~26000 MACs — minor).

## Two other gaps that block sole-engine status

**1. One bad creature aborts the whole batch.** Any load/parse/shape failure returns `Err` on the first bad file (`multi_score.rs:182-262`) — exit 1, **no JSON, no partial results**. Verified: one malformed `bad.json` beside a valid creature ⇒ zero scores emitted. For a NEAT generation that means losing all N scores because one creature serialised badly. A per-creature `{"error": "..."}` entry in the result map would be far more usable. This is a contract decision the TS caller must sign off on before its fallback can be deleted.

**2. No early-exit / streaming on the CLI.** `score_from_creature_dir_with_early_exit` (`multi_score.rs:533`) and the `PartialScore` callback exist as a **Rust library API only** — `--help` exposes no flag. Blocks mid-training pruning of hopeless creatures from the TS side.

Lesser: no incremental/resumable scoring or daemon mode (every invocation re-streams the whole corpus); no per-creature timeout; `--creature-stdin` is single-creature only; directory mode requires identical `input`/`output` across all creatures (`multi_score.rs:249-262`).

## Bonus finding: `--gpu auto` never uses the GPU for production-scale creatures

The consuming fleet runs ~3,000-neuron / ~26,000-synapse creatures. The 256-neuron cap is no longer a rejection (`forward_mse_batched.rs:462` routes >256 to the runtime-sized `forward_mse_scratch` kernel), but `gpu/mod.rs:278-300` declines the GPU on performance grounds for deep pools:

```rust
DirectoryGpuTopology::AllPrivate => true,
DirectoryGpuTopology::ScratchOnly if probe.shallow => true,   // #467
DirectoryGpuTopology::Mixed | DirectoryGpuTopology::ScratchOnly => false, // #317
```

"Shallow" means ≤256 **non-input** neurons, so production creatures classify deep-ScratchOnly ⇒ **CPU always**. Confirmed on a 2901-neuron / 26150-synapse creature: `[gpu] auto fallback to CPU directory mode: deep scratch-kernel creatures are faster on CPU`.

I measured this from the consumer side and it is worth recording, because it makes benchmarks easy to get wrong. On the same 2,266,178-record corpus:

| run | backend | wall | CPU |
|---|---|---|---|
| 1 × 18-neuron creature | **metal** | 13.84 s | 1.4 s |
| 20 × mixed, incl. 3012-neuron | **cpu-fallback** | 126.8 s | 946 s |

Those two are not comparable — different backend *and* different creature size. Re-measured properly, 20 **distinct** ~3,000-neuron champions, all CPU:

| creatures | wall | CPU | wall/creature |
|---|---|---|---|
| 1 | 10.05 s | 55.5 s | 10.05 s |
| 5 | 39.75 s | 293.8 s | 7.95 s |
| 20 | 155.67 s | 1237.7 s | 7.79 s |

So directory mode does amortise the corpus read (`multi_score.rs:835` is the only `for_each_read_chunk`, guarded by `tests/single_pass_assertion.rs`), but per-creature compute dominates: batching saves ~23% versus scoring separately, not orders of magnitude. Also worth noting a **Mixed** pool — one small creature among big ones — drops the *whole batch* to CPU (`gpu/mod.rs:292`).

Question for this repo: is the `Mixed ⇒ CPU` rule intended, and is the deep-ScratchOnly heuristic still right? For a single 18-neuron creature the GPU was *slower* than CPU (13.84 s vs 6.89 s), which is consistent with #317's conclusion.

## Open questions

1. What fraction of a real population is `forwardOnly=false`? If common, a batched recurrent kernel is worth filing separately.
2. **Do TS and Rust agree on recurrent semantics?** Rust's `forwardOnly=false` resets hidden state before every record (`loss.rs:345-349`) — stateless per row, a back edge reads `0.0`. If the TS/WASM path carries state across rows, deleting it changes scores for recurrent creatures. (Related: NEAT-AI honours a `feedbackLoop` option that Rust has no equivalent for.)
3. Should a bad creature fail the batch or be reported per-creature?
4. Is there any test proving `rust_scorer <dir>` and `rust_scorer <file>` produce identical errors for the same creature? I found none — if they can differ, collapsing to one engine just moves the disagreement inside this repo.

Related: stSoftwareAU/NEAT-AI#3853 (RMSE disagreement between the engines), and the GRQ-side fallback removal.

🤖 Raised from a fleet investigation (GRQ #4326 / #4337).

stservice:
# PR Summary — Issue #579

## Summary

Directory mode now scores `forwardOnly: false` (recurrent) creatures instead of
refusing to load them. Closes #579.

The guard in `load_creatures_from_dir` was a correctness stopgap for a hard-coded
boolean: the per-chunk hot loop passed a literal `true` to
`cost::accumulate_cost_sum`, so without the guard a recurrent creature would have
been scored with network state leaking across records — silently wrong. The flag
is now threaded per worker, and the guard is gone.

- **`rust_scorer/src/multi_score.rs`**
  - Deleted the load-time `forwardOnly=false` rejection.
  - Built `worker_forward_only` alongside the existing `worker_creature_idx`, so
    the `par_iter` closure indexes the owning creature's flag with the
    `worker_idx` it already carries. The empty-chunk cost-dispatch probe now uses
    worker 0's flag rather than a literal `true`.
  - Both result builders (CPU and GPU) report the creature's **own**
    `forwardOnly`, not the batch's former precondition.
  - Module docs gained a "Recurrent creatures" section; the
    `gpu_directory_compatible` doc no longer cites `forwardOnly=false` as a load
    failure.
- **Docs** — README directory-mode bullet, a new
  [Recurrent creatures in directory mode](../../../README.md#recurrent-creatures-in-directory-mode-issue-579)
  section with a Mermaid dispatch diagram, the cost-dispatch note, and a
  `CHANGELOG.md` entry.

Everything downstream was already flag-driven, so nothing else needed to change:

- `cost::accumulate_cost_sum` forwards `forward_only` to
  `neat_core::loss::*_sum_batch_packed` for all eight costs.
- With `forward_only = false` the upstream `packed_record_scan` calls
  `CompiledNetwork::reset_state()` before every record, so records stay
  **independent** — which is what keeps partitioning a chunk across Rayon workers
  valid for a recurrent creature. No restructuring of the worker-split machinery.
- **The GPU path needs no change** (issue item 3, confirmed by reading the code):
  `forward_mse_batched.wgsl` and `forward_mse_scratch.wgsl` zero every non-input
  activation per `(creature, record)` thread, so a GPU thread never carries state
  between records; `gpu_directory_probe_for_dir` and `gpu_directory_compatible`
  classify neuron counts and squashes only, never `forwardOnly`.

**Expected cost:** recurrent creatures lose the upstream 8-way/4-way SIMD batch
kernels (both gated on `forward_only`) and fall to the scalar `packed_record_scan`
— several times slower **for those creatures only**. Forward-only creatures in the
same batch are untouched, and `reset_state()` is O(num_neurons) stores per record.
This is a capability change, not a performance change, so no before/after
benchmark applies.

## Evidence

Backend/CLI change — there is no web interface to screenshot. The evidence is the
test suite plus the dispatch change itself.

```mermaid
flowchart LR
    L["load_creatures_from_dir()<br/>#579: no forwardOnly guard"] --> W["worker pool<br/>worker → (creature, forwardOnly)"]
    W --> C["accumulate_cost_sum(..., worker_forward_only[worker_idx])"]
    C -->|true| S["SIMD 8-way / 4-way batch<br/>state carried (no back edges)"]
    C -->|false| R["scalar packed_record_scan<br/>reset_state() before each record"]
    S --> F["per-creature error"]
    R --> F
```

Full local gate results:

- `cargo clippy --workspace --all-targets --all-features -- -D warnings …` — clean.
- `cargo check`, `cargo build`, `RUSTDOCFLAGS=-D warnings cargo doc`,
  `cargo build --release` — all clean.
- `cargo test --workspace --all-features -- --test-threads=2` — every suite
  passes (285 unit tests, 32 doc-tests, and every integration suite including the
  three new recurrent tests).
- `markdownlint-cli2` — 0 issues.
- **`./quality.sh` cannot complete in this container**: its codespell preflight
  exits before the Rust steps with `spell-check: codespell is not installed.`, and
  the image has no `pip`/`ensurepip` to install it (`/usr/bin/python3: No module
  named pip`). Every other gate step was run individually, in `quality.sh`'s own
  order, with the results above. The codespell CI job still gates the PR.

Answering the issue's open question 4 ("is there any test proving
`rust_scorer <creatures_dir>` and `rust_scorer <creature.json>` produce identical
errors for the same creature?"): there is now —
`recurrent_directory_tdd::directory_mode_recurrent_score_matches_single_creature_mode`
compares the two paths' `error` and `score` **bit-for-bit**.

## Test Plan

Added — `rust_scorer/tests/recurrent_directory_tdd.rs` (new file; all fixtures use
exact binary fractions so every partial sum is exact in `f64`, making the
comparisons independent of how a chunk is partitioned across workers):

- `directory_mode_scores_forward_only_false` — a directory holding a creature with
  a genuine back edge exits 0, emits JSON, reports `forwardOnly: false`, scores
  every record, and produces the **reset-semantics** error (`0.1171875`, the
  analytic value for `output = 0.5 × input`). A leaked activation gives a
  different number, so this is the regression test for the hard-coded `true`.
- `directory_mode_recurrent_score_matches_single_creature_mode` — directory mode
  and single-creature mode agree bit-for-bit on `error` and `score`.
- `directory_mode_mixes_forward_only_and_recurrent_creatures` — a **mixed** batch:
  the recurrent creature keeps reset semantics, the forward-only creature beside
  it is unaffected, each entry reports its own `forwardOnly`, and both match their
  single-creature runs.

Added — `rust_scorer/tests/cost_parity.rs`:

- `recurrent_chunk_dispatch_matches_per_record_dispatch_for_every_cost` — for all
  eight costs, handing a whole chunk to `accumulate_cost_sum` with
  `forward_only = false` (what directory mode does) is bit-identical to the
  one-record-at-a-time loop the single-creature path runs. Nine records, so the
  fixture crosses the SIMD-batch boundary the forward-only paths use.
- `recurrent_flag_changes_the_chunk_sum_on_a_back_edge` — the flag is load-bearing:
  on a creature with a back edge, skipping the reset changes the sum, and the
  reset value matches the analytic expectation.

Modified — `rust_scorer/tests/directory_mode_tdd.rs`:

- `directory_mode_rejects_forward_only_false` → `directory_mode_accepts_forward_only_false`.
  **This is a deliberate business-logic change**: the old test pinned the exact
  behaviour the issue asks to remove, so it could not be kept as-is. The
  replacement holds the load-time contract (exit 0, real JSON, the creature's own
  flag echoed back) and its doc comment records the swap.

## Out of scope

The issue also lists three items beyond its "Proposed work", left untouched here
and unchanged by this PR:

- **One bad creature aborts the whole batch** — a per-creature `{"error": …}` entry
  is a contract change the TypeScript caller must sign off on first.
- **No CLI early-exit / streaming flag** — `score_from_creature_dir_with_early_exit`
  remains a library-only API.
- **`--gpu auto` declines the GPU for deep ScratchOnly and Mixed pools** — that
  routing is Issue #317/#467 policy backed by fleet benchmarks, and revisiting it
  needs its own before/after evidence.

---

🤖 Processed by: stservice
🏷️ Run id: `vibe-mt65783r-bf9346`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,28 @@ section to the released version with its date.
 
 ### Added
 
+- **Directory mode scores `forwardOnly: false` (recurrent) creatures
+  (Issue #579).** The load-time guard in
+  `rust_scorer/src/multi_score.rs` that rejected any creature directory
+  containing a recurrent creature is gone. The per-chunk worker pool now carries
+  a `worker_forward_only` vector alongside `worker_creature_idx`, so each worker
+  passes **its own creature's** flag to `cost::accumulate_cost_sum` instead of a
+  hard-coded `true`; with `forward_only = false` the upstream kernel clears the
+  network state before every record, which keeps records independent and
+  therefore keeps the Rayon record partition valid. A directory may mix
+  forward-only and recurrent creatures, each entry's `forwardOnly` JSON field
+  now reports the creature's own flag, and `rust_scorer <dir>` reports the same
+  `error` as `rust_scorer <file>` for the same recurrent creature. The GPU path
+  needed no shader change — both kernels already zero every non-input activation
+  per `(creature, record)` thread. Recurrent creatures lose the upstream SIMD
+  batch kernels (gated on `forward_only`) and fall to the scalar per-record
+  scan; forward-only creatures in the same batch are unaffected. New coverage:
+  `rust_scorer/tests/recurrent_directory_tdd.rs` (reset semantics, single-creature
+  parity, mixed batch) plus two `rust_scorer/tests/cost_parity.rs` tests
+  (chunk-vs-per-record dispatch bit-identical for all eight costs; the
+  `forward_only` flag is load-bearing on a back edge). The pre-#579
+  `directory_mode_rejects_forward_only_false` test is replaced by
+  `directory_mode_accepts_forward_only_false` — a deliberate behaviour change.
 - **Binary-level regression gate for creatures carrying a `memetic` block
   (stSoftwareAU/NEAT-AI#3813).** `rust_scorer/tests/memetic_creature_parse.rs`
   feeds two committed fixtures — `rust_scorer/tests/fixtures/memetic_creature.json`
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -750,7 +750,7 @@ dependencies = [
 
 [[package]]
 name = "neat-core"
-version = "0.10.2"
+version = "0.10.5"
 dependencies = [
  "serde",
  "serde_json",
@@ -1076,7 +1076,7 @@ checksum = "19b30a45b0cd0bcca8037f3d0dc3421eaf95327a17cad11964fb8179b4fc4832"
 
 [[package]]
 name = "rust_scorer"
-version = "1.1.67"
+version = "1.1.68"
 dependencies = [
  "bytemuck",
  "clap",
--- a/README.md
+++ b/README.md
@@ -133,7 +133,46 @@ rust_scorer <creature.json | creatures_dir> <training_data_dir>
 
 - `creature.json` path: scores one creature and returns the existing single-object output.
 - `creatures_dir` path: scores every `*.json` in that directory in one pass over training data and returns one JSON object keyed by each file's stem (filename without extension or folders).
-- Directory mode requires `forwardOnly: true` and matching `input` / `output` shape across all files.
+- Directory mode requires a matching `input` / `output` shape across all files. Since **Issue #579** it accepts `forwardOnly: false` creatures — see [Recurrent creatures in directory mode](#recurrent-creatures-in-directory-mode-issue-579) below.
+
+### Recurrent creatures in directory mode (Issue #579)
+
+A creature directory may mix `forwardOnly: true` and `forwardOnly: false`
+creatures. Each is scored under **its own** flag, so a recurrent creature in a
+batch gets the same per-record state reset the single-creature path applies —
+`neat_core::loss::packed_record_scan` calls `CompiledNetwork::reset_state()`
+before every record, a back edge therefore reads `0.0`, and records stay
+independent. Independence is what keeps the batch path valid: a chunk is still
+partitioned across Rayon workers, and each creature's `forwardOnly` follows its
+worker.
+
+```mermaid
+flowchart LR
+    L["load_creatures_from_dir()"] --> W["worker pool<br/>(worker → creature, forwardOnly)"]
+    W --> C["accumulate_cost_sum(..., forward_only)"]
+    C -->|true| S["SIMD 8-way / 4-way batch<br/>(state carried — no back edges)"]
+    C -->|false| R["scalar packed_record_scan<br/>reset_state() per record"]
+    S --> F["per-creature error"]
+    R --> F
+```
+
+Consequences worth knowing before you batch a recurrent population:
+
+- **Same answer as single-creature mode.** `rust_scorer <dir>` and
+  `rust_scorer <file>` report the same `error` for the same recurrent creature
+  (pinned by `rust_scorer/tests/recurrent_directory_tdd.rs`).
+- **`forwardOnly` is echoed per creature** in the directory JSON, so a caller
+  can see which entries took the recurrent path.
+- **Recurrent creatures lose the SIMD batch kernels** (upstream gates the 8-way
+  and 4-way paths on `forward_only`) and fall to the scalar per-record scan —
+  several times slower *for those creatures only*. Forward-only creatures
+  sharing the batch are unaffected; `reset_state()` costs O(neurons) stores per
+  record.
+- **The GPU path needs no change.** Both `forward_mse_batched` and
+  `forward_mse_scratch` zero every non-input activation per
+  `(creature, record)` thread, so a GPU thread never carries state between
+  records — the reset semantics hold on GPU by construction, and the
+  hostability/topology probes classify neuron counts only.
 
 ### GPU mode (Issues #80 / #83)
 
@@ -1104,7 +1143,10 @@ forward-only path and the per-record recurrent path:
 - **Per-record recurrent path.** `forwardOnly: false` creatures use
   `TrainingDataIterator` to feed `[inputs..., targets...]` packed records into
   the same `accumulate_cost_sum` helper one record at a time, so every
-  supported cost works here too.
+  supported cost works here too. Directory mode (Issue #579) passes a whole
+  chunk to the same helper with `forward_only = false`; the per-record reset
+  inside `packed_record_scan` makes the two numerically identical, pinned per
+  cost by `rust_scorer/tests/cost_parity.rs`.
 - **`CATEGORICAL_ERROR` dispatches** through `categorical_error_sum_batch_packed`
   (landed via [`NEAT-AI-core#88`](https://github.com/stSoftwareAU/NEAT-AI-core/issues/88);
   unblocked here in #134) — the dispatch returns the integer count of
--- a/docs/archive/pr-summaries/pr-summary-579.md
+++ b/docs/archive/pr-summaries/pr-summary-579.md
@@ -1,0 +1,135 @@
+# PR Summary — Issue #579
+
+## Summary
+
+Directory mode now scores `forwardOnly: false` (recurrent) creatures instead of
+refusing to load them. Closes #579.
+
+The guard in `load_creatures_from_dir` was a correctness stopgap for a hard-coded
+boolean: the per-chunk hot loop passed a literal `true` to
+`cost::accumulate_cost_sum`, so without the guard a recurrent creature would have
+been scored with network state leaking across records — silently wrong. The flag
+is now threaded per worker, and the guard is gone.
+
+- **`rust_scorer/src/multi_score.rs`**
+  - Deleted the load-time `forwardOnly=false` rejection.
+  - Built `worker_forward_only` alongside the existing `worker_creature_idx`, so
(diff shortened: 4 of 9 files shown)


stservice:

Last 20 lines of its output:
  bump: cc -> 1.4.4
  bump: clap -> 4.6.6
  bump: clap_builder -> 4.6.6
  bump: either -> 1.18.0
  bump: futures-core -> 0.3.34
  bump: futures-task -> 0.3.34
  bump: futures-util -> 0.3.34
  bump: log -> 0.4.34
  bump: ordered-float -> 5.5.0
  bump: pkg-config -> 0.3.34
  fail: syn -> 2.0.119 (cargo update rejected)
  bump: thiserror -> 2.0.20
  bump: xml-rs -> 0.8.29
  bump: zerocopy -> 0.8.56
external: 19 bumped, 0 deferred, 1 failed

- `Cargo.lock`

```
  bump: cc -> 1.4.4
  bump: clap -> 4.6.6
  bump: clap_builder -> 4.6.6
  bump: either -> 1.18.0
  bump: futures-core -> 0.3.34
  bump: futures-task -> 0.3.34
  bump: futures-util -> 0.3.34
  bump: log -> 0.4.34
  bump: ordered-float -> 5.5.0
  bump: pkg-config -> 0.3.34
  fail: syn -> 2.0.119 (cargo update rejected)
  bump: thiserror -> 2.0.20
  bump: xml-rs -> 0.8.29
  bump: zerocopy -> 0.8.56
external: 19 bumped, 0 deferred, 1 failed

```

</details>

Issue #581, opened 2026-08-24
Pin cross-engine parity for (from, to, type)-keyed synapses before the rule relaxes


nleck:
Scorer half of [NEAT-AI#3873](https://github.com/stSoftwareAU/NEAT-AI/issues/3873)
and [NEAT-AI-core#577](https://github.com/stSoftwareAU/NEAT-AI-core/issues/577),
which propose keying synapses by `(from, to, type)` so one source can feed both
branches of an `IF` without an intermediate IDENTITY relay.

## Why this repository cares

This is the engine that would *disagree* first. Today a creature carrying
`A→B positive` and `A→B negative`:

- **TypeScript** rejects the second synapse — `connect()` asserts
  `"Connection already exists"`, keyed on `(from, to)` alone;
- **`rust_scorer`** sums both.

So such a creature means two different things depending on who scores it. That
divergence has already produced a production "improvement" that was not real,
which is why `if_tree_fixture.rs` and `shallow_fixture.rs` carry regression
guards about duplicate pairs today.

If the rule is relaxed upstream, this engine's behaviour is already the correct
one — but that must be *pinned*, not assumed.

## What to do here

- Once neat-core lands the rule, take the version bump and confirm nothing in
  the scorer keys synapses by `(from, to)` on its own.
- Add a cross-engine parity fixture: an `IF`-output creature where one source
  feeds the same target as both `positive` and `negative`, scored by
  `rust_scorer` and by NEAT-AI's `Creature.scoreDir`, asserting the scores agree
  and that TypeScript drops no synapse on load (the existing parity harness in
  NEAT-AI-Forests' `ts_parity.rs` already checks `jsonSynapses ===
  loadedSynapses` — the same assertion is what would have caught the original
  divergence).
- Update the fixture comments that currently state the `(from, to)` rule as
  fact.

## Not urgent

Nothing breaks until the upstream rule changes; every creature valid today
stays valid. This issue exists so the parity guard lands with the change rather
than after the first creature that needs it.

stservice:
✅ **Vibe Coder released this claim** — host `vibe-coder-31411`, finished 07:35 UTC. Raised #582 — https://github.com/stSoftwareAU/NEAT-AI-scorer/pull/582
**Attempts on this issue:** 2
- 04:33 `vibe-coder-34648` — no PR (`timeout`, phase `execute`)
- 07:35 `vibe-coder-31411` — raised #582

stservice:
[...]

PR #587, opened 2026-08-26
dual_role_parity: bit-exact score equality between the relay-free and relay forms fails on x86-64 (2 ULP) (Issue #585)

### LINKED ISSUE #585: dual_role_parity: bit-exact score equality between the relay-free and relay forms fails on x86-64 (2 ULP)
## Summary

`tests/dual_role_parity.rs::directory_scoring_agrees_between_the_forms_and_separates_the_dropped_one`
fails on an x86-64 Linux host at `HEAD` of `Develop`, independently of any
in-flight change. It asserts **bit-exact** equality between the relay-free
creature and its relay workaround; the two differ by 2 ULP:

```text
thread '...' panicked at rust_scorer/tests/dual_role_parity.rs:286:5:
assertion `left == right` failed: the relay-free creature and its relay workaround must score identically
  left: 1.3229378675896557
 right: 1.3229378675896573
```

## What was checked

* Reproduced on a clean checkout of `Develop` (no working-tree changes),
  toolchain 1.98.0, sibling `NEAT-AI-core` at 0.10.6.
* Not a reader-count / reduction-order artefact: identical values under
  `NEAT_SCORER_FILE_THREADS=1`, `=2` and `=4`.
* The other nine tests in the file pass, including the CPU-activation
  reference and both GPU parity tests.

The relay form inserts an extra `IDENTITY` neuron, so the two shapes sum the
same contributions in a different association order. On the host the Issue #581
contract was pinned on, that happened to land on the same bits; on x86-64 it
does not.

## What a fix looks like

Decide what the #581 contract actually requires and encode it: either the two
forms must be bit-identical on every target (which needs the summation order
made identical, not merely equivalent), or *scores* agree within a stated
tolerance while the bit-exact requirement stays where it is genuinely testable
(the CPU-vs-independent-reference assertion). The README
"IF decision-tree parity contract" section and this assertion must say the same
thing.

Found while working #583; unrelated to that change, so filed separately rather
than folded into it.

stservice:
# Bound the dual-role parity score comparison to the reduction it can hold (Issue #585)

## Summary

`tests/dual_role_parity.rs::directory_scoring_agrees_between_the_forms_and_separates_the_dropped_one`
asserted **bit-exact** equality between the relay-free creature's
directory-pipeline score and its relay workaround's, and failed by 2 ULP
(`1.3229378675896557` vs `1.3229378675896573`). Closes #585.

The reported cause — x86-64, or the relay's extra `IDENTITY` neuron changing the
association order of the creature's own arithmetic — is not the cause. Directory
scoring folds each creature's chunk from **as many `f64` partial sums as that
creature was allotted workers**, and `multi_score::workers_per_creature_split`
allots a *ragged* count when `activation_threads` is not a multiple of the
population (3 creatures on 8 threads → `[3, 3, 2]`, Issue #537). The three
fixtures sort as `dropped`, `relaxed`, `relayed`, so `relaxed` reduces
`683 + 683 + 682` records while `relayed` reduces `1024 + 1024` — the same
per-record errors, a different association order, the last bits of an `f64` sum
apart. Nothing architecture-specific: the reported failure reproduces
bit-for-bit on this **aarch64** host under `NEAT_SCORER_ACTIVATION_THREADS=8`,
and disappears at `6`, `7` or `12`.

Encoding the contract (the issue's option 2 — a stated tolerance where the
requirement is not genuinely testable, bit-exactness where it is):

| Comparison | Bound |
|---|---|
| CPU activation vs the independent reference | bit-exact (unchanged) |
| relay-free vs relay activation, per record | bit-exact (unchanged) |
| relay-free vs relay whole-corpus loss, one partition | bit-exact (**new test**) |
| relay-free vs relay **directory-pipeline** score | `1e-12` relative (**was `==`**) |
| CPU vs GPU per-creature loss | `1e-3` relative (unchanged) |

Option 1 (forcing the two forms to reduce in an identical order) was rejected:
it would require every creature in a directory to receive the same worker count,
idling up to `n_creatures - 1` threads whenever the population is smaller than
the thread budget — a throughput regression to buy a property the new bit-exact
test already pins where it is real. No production scoring code changed.

## Evidence

Backend/CLI change with no web interface, so no screenshot: the evidence is the
reproduction and the gate.

**Reproduced, then fixed** — `directory_scoring_agrees_…` swept across activation
thread counts, before and after:

| `NEAT_SCORER_ACTIVATION_THREADS` | worker allotment | before | after |
|---|---|---|---|
| 3, 6, 12 | `[1,1,1]` / `[2,2,2]` / `[4,4,4]` | pass | pass |
| 5 | `[2,2,1]` | **FAIL** (`…6573` vs `…6493`) | pass |
| 7 | `[3,2,2]` | pass | pass |
| 8 | `[3,3,2]` | **FAIL** (`…6557` vs `…6573` — the issue's exact values) | pass |

Measured relative differences at the two failing counts:

```text
threads=5: parity_drift=6.04e-15  dropped_gap=9.01e-3
threads=8: parity_drift=1.17e-15  dropped_gap=9.01e-3
```

The `1e-12` bound sits ~165× above the worst drift and ~2× above the
`n_records × f64::EPSILON` ≈ `4.5e-13` re-association bound for this
2,048-record corpus, while a dropped branch edge — the divergence the guard
exists to catch — stays `9.0e-3` away, ten orders of magnitude clear.

```mermaid
flowchart LR
    R["2,048 records<br/>identical per-record errors"]
    R --> A["relaxed — 3 workers<br/>683 + 683 + 682"]
    R --> B["relayed — 2 workers<br/>1024 + 1024"]
    A --> SA["Σ f64 → 1.3229378675896557"]
    B --> SB["Σ f64 → 1.3229378675896573"]
    SA --> D["2 ULP apart — reduction noise,<br/>not a difference in the function"]
    SB --> D
    D --> T["bit-exact where the order is fixed;<br/>1e-12 where the pipeline may re-associate"]
```

**Quality gate.** `./quality.sh < /dev/null` runs clean up to the `codespell`
preflight, which cannot run on this container — `codespell` is not installed and
the image has no `pip`/`ensurepip` (`/usr/bin/python3: No module named pip`).
Every remaining step was then run individually and passed: `cargo fmt --all`,
`cargo clippy --workspace --all-targets --all-features -D warnings`,
`cargo check`, `cargo build --workspace`,
`cargo test --workspace --all-features -- --test-threads=2` (0 failures),
`RUSTDOCFLAGS=-D warnings cargo doc --workspace --no-deps`,
`cargo build --workspace --release`, plus `markdownlint-cli2` (0 issues in 188
files). CI runs the spell-check job for real.

## Test Plan

Added:

- `rust_scorer/tests/dual_role_parity.rs::the_whole_corpus_loss_is_bit_identical_between_the_forms`
  — reduces all 2,048 records through one `mse_sum_batch_packed` call per form
  and asserts the relay-free and relay losses are equal **to the bit**, with the
  dropped-edge creature differing. This is where "the relay changes nothing" is
  genuinely testable, and it is the assertion that would fail if the relaxed
  shape were ever mis-scored.
- `rust_scorer/src/multi_score.rs::tests::a_ragged_worker_allotment_partitions_the_same_chunk_differently`
  — calls `workers_per_creature_split(3, 8, 1, usize::MAX)` and
  `partition_packed_record_ranges` and asserts the same chunk is split into a
  different number of sub-ranges, each still tiling the whole buffer. Pins the
  mechanism as a fact about the partition rather than an observation about one
  host.

Modified (documented behaviour change):

- `directory_scoring_agrees_between_the_forms_and_separates_the_dropped_one` —
  the relaxed-vs-relayed comparison is now `relative_difference ≤ 1e-12`
  (`CPU_PIPELINE_REL_TOLERANCE`) instead of `==`, and the dropped-creature
  comparison is strengthened from `assert_ne!` to "differs by more than `1e6 ×`
  the tolerance", so loosening one end cannot quietly weaken the other. No test
  was removed or commented out.

Docs updated to state the same contract: the README
"Synapses are keyed by `(from, to, type)`" section (bound table, mechanism,
Mermaid diagram), the `dual_role_fixture.rs` module and
`relay_equivalent_if_creature_json` doc comments (which claimed the score
equality was "bit-exact and not a tolerance"), the `dual_role_parity.rs` module
"Documented tolerance" paragraph, and the CHANGELOG (new `Fixed` entry; the
Issue #581 `Added` entry's "score identically" wording corrected).

---

🤖 Processed by: stservice
🏷️ Run id: `vibe-mt9csrw5-2f7b02`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,8 +32,9 @@ section to the released version with its date.
   is identical in the raw JSON, the parsed export and the compiled network — the
   Rust-side spelling of the `jsonSynapses === loadedSynapses` assertion
   `NEAT-AI-Forests`' `ts_parity.rs` makes against `Creature.scoreDir` — that the
-  relaxed and relay forms activate bit-identically and score identically through
-  the real directory pipeline, that the dropped-edge creature scores
+  relaxed and relay forms activate bit-identically and score equal through
+  the real directory pipeline (see the Issue #585 entry below for the exact
+  bound each comparison carries), that the dropped-edge creature scores
   *differently* (so the guard cannot pass vacuously), and that both GPU kernels
   agree with CPU within the `1e-3` cross-backend tolerance. Also new:
   `fixture_json::constant_neuron_json`, which omits the `"squash"` key that
@@ -187,6 +188,33 @@ section to the released version with its date.
   `docs/performance-baseline.md`.
 
 ### Fixed
+
+- **`dual_role_parity` no longer demands bit-exact equality from a reduction
+  that is free to re-associate (Issue #585).**
+  `directory_scoring_agrees_between_the_forms_and_separates_the_dropped_one`
+  asserted `==` between the relay-free creature's directory-pipeline score and
+  its relay workaround's, and failed by 2 ULP
+  (`1.3229378675896557` vs `…73`). The cause is not the architecture it was
+  reported on and not the extra `IDENTITY` relay: directory scoring folds each
+  creature's chunk from as many `f64` partial sums as that creature was allotted
+  workers, and `workers_per_creature_split` allots a **ragged** count when
+  `activation_threads` is not a multiple of the population (3 creatures on 8
+  threads → `[3, 3, 2]`, Issue #537). Two creatures then reduce the identical
+  per-record errors in a different association order. The reported failure
+  reproduces bit-for-bit under `NEAT_SCORER_ACTIVATION_THREADS=8` on aarch64 and
+  disappears at `6`, `7` or `12`. The parity contract now states a bound per
+  comparison: **bit-exact** for every per-record activation and for the
+  whole-corpus loss reduced in one partition (new
+  `the_whole_corpus_loss_is_bit_identical_between_the_forms`, which calls
+  `mse_sum_batch_packed` over all 2,048 records), **`1e-12` relative** for the
+  directory-pipeline score, and the unchanged **`1e-3` relative** cross-backend
+  tolerance. Measured drift across thread counts `1..=16` is ≤ `6.1e-15`; a
+  dropped branch edge still moves the loss by `9.0e-3`, so the guard separates a
+  real divergence by ten orders of magnitude. New
+  `multi_score::tests::a_ragged_worker_allotment_partitions_the_same_chunk_differently`
+  pins the mechanism, and the README
+  "Synapses are keyed by `(from, to, type)`" section carries the bound table.
+  No production scoring code changed.
 
 - **A `wgpu` device lost mid-run falls back to CPU instead of panicking
   (Issue #583).** `wgpu` reports a lost device *fatally* — `Device::poll` calls
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1076,7 +1076,7 @@ checksum = "19b30a45b0cd0bcca8037f3d0dc3421eaf95327a17cad11964fb8179b4fc4832"
 
 [[package]]
 name = "rust_scorer"
-version = "1.1.71"
+version = "1.1.72"
 dependencies = [
  "bytemuck",
  "clap",
--- a/README.md
+++ b/README.md
@@ -484,13 +484,52 @@ asserts that:
   JSON, the parsed export and the compiled network (the Rust-side spelling of
   the `jsonSynapses === loadedSynapses` assertion `NEAT-AI-Forests`' `ts_parity.rs`
   makes against `Creature.scoreDir`);
-- **the relaxed form and the relay workaround agree exactly** — bit-identical
-  activations and an identical loss through the real directory pipeline, so
-  upstream may drop the relay without moving a score;
+- **the relaxed form and the relay workaround agree** — bit-identical
+  activations, a bit-identical whole-corpus loss, and an equal score through the
+  real directory pipeline, so upstream may drop the relay without moving a
+  score;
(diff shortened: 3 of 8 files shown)


stservice:

Last 20 lines of its output:
  bump: cc -> 1.4.4
  bump: clap -> 4.6.6
  bump: clap_builder -> 4.6.6
  bump: either -> 1.18.0
  bump: futures-core -> 0.3.34
  bump: futures-task -> 0.3.34
  bump: futures-util -> 0.3.34
  bump: log -> 0.4.34
  bump: ordered-float -> 5.5.0
  bump: pkg-config -> 0.3.34
  bump: thiserror -> 2.0.20
  bump: xml-rs -> 0.8.29
  bump: zerocopy -> 0.8.56
external: 19 bumped, 0 deferred, 0 failed

- `Cargo.lock`

```
  bump: cc -> 1.4.4
  bump: clap -> 4.6.6
  bump: clap_builder -> 4.6.6
  bump: either -> 1.18.0
  bump: futures-core -> 0.3.34
  bump: futures-task -> 0.3.34
  bump: futures-util -> 0.3.34
  bump: log -> 0.4.34
  bump: ordered-float -> 5.5.0
  bump: pkg-config -> 0.3.34
  bump: thiserror -> 2.0.20
  bump: xml-rs -> 0.8.29
  bump: zerocopy -> 0.8.56
external: 19 bumped, 0 deferred, 0 failed

```

</details>

